PR #4, opened 2022-09-14
Update review BE ci/cd

veidenberg:
This PR updates gitlab/Kustomize config to fix Tools API review branch deployments (adds review-specific labels and selectors to avoid mixup of default/review Tools API pods connected to client review apps). Also includes small cleanups.


k8s/overlays/review/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: ensembl-dev
```
kamaldodiya, line 3: This is implied by the environment so need not repeat here. 
`  environment:
     name: wp-hx-dev`

veidenberg: Didn't know that, thanks! To clarify:
-  What is the purpose of `namespace` key in our kustomization files and why is the value different from the corresponding environment name in gitlab ci/cd config?
-  To be consistent, should the `namespace` key also be removed from all the other kustomization manifests, both in here and other BE repos? 
- Since `namespace` is often the only directive in the subfolder kustomization files (in addition of including the `base` folder), can all of these kustomization files be removed (together with their folders) and the gitlab ci/cd config should then just use the base kustomization in all the deployment jobs in all the environments? Or, is it better to just remove the namespace key and keep the boilerplate kustomization subfolders/files files as placeholders for future expansion?

PR #7, opened 2023-11-06
Update Tools API for 241

veidenberg:
* Remove species UUID=>GCA mapping from Tools API code
* Maintenance (bump versions, update deprecated code, fix tests, add configmap)

Review app: http://test-blast.review.ensembl.org

JIRA:
* https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2280
* https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2066


app/main.py
```diff
@@ -83,33 +80,28 @@ class QuerySequence(BaseModel):
...
     job_ids: list[str]
 
 
-# Infer the target species index file for BLAST payload
-def get_blast_filename(genome_id: str, db_type: str) -> str:
-    suffix = f"{db_type}.toplevel" if db_type.startswith("dna") else f"{db_type}.all"
-    return f"ensembl/{app.genome_ids[genome_id]}/{db_type}/{app.genome_ids[genome_id]}.{suffix}"
+# Infer the target db path for BLAST job
+def get_db_path(genome_id: str, db_type: str) -> str:
+    suffix_map = {"dna": "unmasked", "dna_sm": "softmasked"} #dna/dna_sm/cdna/pep
```
kamaldodiya: Why not global ?

app/main.py
```diff
@@ -83,33 +80,28 @@ class QuerySequence(BaseModel):
...
 
 # Submit a BLAST job to JD. Returns a resolvable for fetching the response.
 async def run_blast(
     query: dict, blast_payload: dict, genome_id: str, db_type: str
 ) -> dict:
     blast_payload["sequence"] = query["value"]
-    blast_payload["database"] = get_blast_filename(genome_id, db_type)
-    url = "http://www.ebi.ac.uk/Tools/services/rest/ncbiblast_ensembl/run"
+    blast_payload["database"] = get_db_path(genome_id, db_type)
+    url = "http://wwwdev.ebi.ac.uk/Tools/services/rest/ncbiblast_ensembl/run"
```
kamaldodiya: I would avoid hardcoded string. Let the config decide based on environment which URL to use

app/main.py
veidenberg: Global `suffix_map`? Why?

kamaldodiya: Every time get_db_path is being called it will assign suffix_map with the dictionary so if you make 10 calls it will be assigned 10 times and used in subsequent step. 
If you keep it global it will be assigned only once and used however many times

app/main.py
veidenberg: I was considering that, but I usually need to tie the URL to a code change rather than to deployment namespace (update on jdispatcher dev => update in toolsapi code => review => dev => prod; at some point live jdispatcher is updated => udpate URL in Tools API). How could I switch configmaps in this case?

kamaldodiya: Updating configmap is way better than frequent changes in the code base itself in my opinion

kamaldodiya: @veidenberg  more on this here https://12factor.net/config

app/main.py
veidenberg: Done

app/main.py
veidenberg: Done (+configmaps applied in k8s)

veidenberg:
Blast jobs work (thanks @ens-st3 for testing), merging PR.

ens-st3:
Tested against mouse GRCm39 and 129S1_SvImJ_v1.
Against genomic gives matches that look correct but GB doesn't work
Against cDNAs gives matches to transcripts that we can look to see which genome they're from

ens-st3:
Also confirmed against softmasked (works) and peptides (IDs of hits are sensible)

PR #10, opened 2024-06-04
Add open api spec for VEP results payload

azangru:
This PR adds an openAPI spec file with the spec for the first endpoint, reporting the results of a VEP analysis.

### Questions for discussion

1. **Should reference allele be shaped the same as alternative allele?**

The biggest question for me is whether it makes sense to model the reference allele and the alternative alleles the same, or they are clearly distinct. For example, the reference allele has a constant `allele_type` (in hypsipyle, it is reported as "biological_region"), which has no value if reference allele is separated into its own field. Similarly, reference allele is not going to have any predicted molecular consequences (i.e. it will always be an empty array).

**DECIDED DURING CODE REVIEW:** Model reference allele and the alternative allele as two distinct objects. I have confirmed with the Variation team that reference alleles do not have predicted molecular consequences.

2. **Is it worth modelling variant's location through the typical slice shape?**

Should the location information be kept inside a nested `slice: { region: {name}, location: {start,end} }` structure, or does it make more sense to just have these fields (`region_name`, `start`, `end`) at the top level?

**DECIDED DURING CODE REVIEW:** Use a simplified notation. Most members of the team preferred `{ location: { region_name, start, end } }` over having these three fields directly at the top level of a variant.

3. **Does variant have allele type?**

The [XD](https://xd.adobe.com/view/2e253ea8-6bd8-4efa-ad39-dc65f9eb01bd-7b76/screen/802e46d2-1597-4821-ad78-20b392e5d925/) shows the allele type string under the variant name in the variant (left) column; but the VEP output produces allele types on per-allele level.

**OUTCOME:**  Could not reach a clear decision on this one. Andrea wants to display allele type at the variant level. I am leaving this field in the variant model for now.

### Jira ticket
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2588

### DIFF
--- a/APISpecification.yaml
+++ b/APISpecification.yaml
@@ -128,7 +128,6 @@ components:
             - region_name
             - start
             - end
-
         reference_allele:
           $ref: '#/components/schemas/ReferenceVariantAllele'
         alternative_alleles:
--- a/APISpecification.yaml
+++ b/APISpecification.yaml
@@ -33,6 +33,24 @@ paths:
                 $ref: '#/components/schemas/VepResultsResponse'
 components:
   schemas:
+    AlternativeVariantAllele:
+      type: object
+      required:
+        - allele_sequence
+        - allele_type
+        - predicted_molecular_consequences
+      properties:
+        allele_sequence:
+          type: string
+        allele_type:
+          type: string
+        representative_population_allele_frequency:
+          type: number
+          nullable: true
+        predicted_molecular_consequences:
+          type: array
+          items:
+            $ref: '#/components/schemas/PredictedTranscriptConsequence'
     PaginationMetadata:
       type: object
       required:
@@ -77,6 +95,13 @@ components:
           enum:
             - forward
             - reverse
+    ReferenceVariantAllele:
+      type: object
+      required:
+        - allele_sequence
+      properties:
+        allele_sequence:
+          type: string
     Variant:
       type: object
       required:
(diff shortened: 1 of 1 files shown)


APISpecification.yaml
```diff
@@ -0,0 +1,157 @@
+openapi: 3.0.0
+info:
+  title: Tools api
+  version: 0.0.1
+paths:
+  /api/tools/vep/submissions/{id}:
```
kamaldodiya: Does the name justify the action ? 
`/vep/submissions/{id}` : I would interpret as you want to get some information about the vep job
If the intention is to get the results then something like `/vep/results/{id}` would be more appropriate

APISpecification.yaml
```diff
@@ -0,0 +1,157 @@
...
+  schemas:
+    PaginationMetadata:
+      type: object
+      required:
+        - page
+        - per_page
+        - total
+      properties:
+        page:
+          type: number
```
kamaldodiya, line 62: Not absolutely necessary but adding example here would populate the example in swagger with real values instead of the type.

kamaldodiya (commented):
Looks good to me. Just one

APISpecification.yaml
```diff
@@ -0,0 +1,157 @@
...
+            type: number
+            default: 1
+        - name: per_page
+          in: query
+          required: false
+          schema:
+            type: number
+            default: 100
+      responses:
+        '200':
```
veidenberg, line 28: Does client expect 404 when results ID is not found?

APISpecification.yaml
azangru: 1) User submits data for a VEP job and receives back a submission id (the missing submit endpoint)
2) User calls the api with the submission id to learn about the status of the submission (the missing status endpoint)
3) Finally, VEP analysis has finished, and user calls the api with the submission id to get the results.

I do not think `/vep/results/{id}` is great; I think `/vep/submissions/{id}` or `/vep/jobs/{id}` is more appropriate (aka restful). But ultimately, it is up to you to decide.

APISpecification.yaml
azangru: Yes; but going back to our discussion during code review today, isn't this standard behaviour? For example, I am sure client can also send an invalid payload (e.g. request a minus first page); but we don't expect the submitter of the spec to describe the 422 error?

APISpecification.yaml
veidenberg: I agree with Kamal, because here you're not expecting a submission instance, but results (i.e. will `submissions` or `jobs` endpoint return results, status, or some other information about the job)? It's also worth considering how will we structure the other endpoint URIs: 
- `POST:vep/run`, ` GET:vep/[status|result]/{id}` (like jDispatcher)
- `POST:vep/job`, `GET:vep/jobs/[status|results]/{id}` (like ToolsAPI BLAST)
- `POST:vep/jobs`, `GET:vep/jobs/{id}/[status|results]` (or `vep/submissions`)

I'd vote for the last one (as [restful](https://stackoverflow.blog/2020/03/02/best-practices-for-rest-api-design/) as we can get).

APISpecification.yaml
veidenberg: Error responses are just like 200 response:
- these need to be implemented on BE side (e.g. specific inspection and handling of payloads from downstream (404) or upstream (422))
- these need to be agreed with and handled/implemented on client side

No problem if we don't include error responses in the spec: it means no need to implement these (for now). Or, if I understand your suggestion correctly, we agree (not in spec but elsewhere) that some error responses (but which?) are always implemented (but using which spec/payloads?), and (for some reason) these are not added to the spec by the spec submitter but by BE service implementer (when?). Either way, it's fine if we skip it now, but error responses still need to be added to the spec at some point (preferrably before implementation).

veidenberg (approved):
200 response payload looks good. Rest of the spec still needs work (URL, existing endpoints, error responses etc.), but we can continue updating this as needed.

kamaldodiya (approved):

azangru:
I am just going to update the url, and to add the field about whether a transcript is canonical... Give me 5 minutes

PR #12, opened 2024-06-11
Vep results module

jonk-ebi:
This PR adds in a VEP module that can parse a VEP results VCF file and output an object compatible with VepResultsResponse defined in APISpecification.yaml

Jira ticket: https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2589

You can test the module using `pytest  app/tests/test_vep.py`

If you want to try the module against a real VCF file then remove the skip decorator from `test_get_results_with_file_and_dump` and update the VCF file path to a local VCF file. Instructions on where to find a suitable VCF file can be found here https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2586


app/vep/model.py
```diff
@@ -0,0 +1,81 @@
...
+class Strand(Enum):
+    forward = 'forward'
+    reverse = 'reverse'
+
+
+class PredictedTranscriptConsequence(BaseModel):
+    feature_type: FeatureType
+    stable_id: str = Field(..., description='transcript stable id, versioned')
+    gene_stable_id: str = Field(..., description='gene stable id, versioned')
+    gene_symbol: str
```
azangru, line 41: How does the model make sure that a field is nullable (e.g. string or null) as opposed to non-nullable (e.g. only string)?

In the spec, we have a bunch of nullable fields, e.g. gene_symbol, or variant.name. If the value coming out of the vcf file is an empty string, I was hoping we could get a null.

app/vep/model.py
```diff
@@ -0,0 +1,81 @@
...
+
+from pydantic import BaseModel, Field
+
+class PaginationMetadata(BaseModel):
+    page: int
+    per_page: int
+    total: int
+
+class PredictedIntergenicConsequence(BaseModel):
+    feature_type: Any = Field(
```
azangru, line 18: This field's value is always a null. The model is being too permissive here.

app/vep/vcf_results.py
```diff
@@ -0,0 +1,235 @@
...
+            allele_type = "indel"
+            so_term = "SO:1000032"
+
+        case [False, False, True]:
+            allele_type = "substitution"
+            so_term = "SO:1000002"
+    return allele_type, so_term
+
+
+def _get_prediction_index_map(
```
kamaldodiya, line 54: Does this remains fixed or changes based on the output vcf ?

app/vep/model.py
jonk-ebi: At the moment it will pass through the empty string. I will change it

kamaldodiya (approved):

app/vep/model.py
jonk-ebi: Change made

app/vep/model.py
jonk-ebi: Currently providing Null in the code. Will look at the model generator to see how to support this

app/vep/vcf_results.py
jonk-ebi: This will change depending on the options provided by the user. Additional columns are added to output the results of different plugins / options

PR #14, opened 2024-06-27
Refactor project to add VEP app with a single file upload endpoint

Mehrnaz-Charkhchi:
- Refactor tools API to add an app for VEP.
- Add sample structure and files for further development.
- provide a sample end point to upload files.
Jira: [ENSWBSITES-2637](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2637)


Dockerfile
```diff
@@ -1,7 +1,9 @@
...
-CMD ["uvicorn", "app.main:app", "--proxy-headers", "--host", "0.0.0.0", "--port", "8013"]
\ No newline at end of file
+WORKDIR /app
+COPY ./requirements.txt requirements.txt
+RUN pip install --no-cache-dir --upgrade -r requirements.txt
+COPY ./app /app
+COPY ./data /data
+ENV PORT 8013
+EXPOSE 8013
+# CMD ["uvicorn", "main:app", "--proxy-headers", "--host", "0.0.0.0", "--port", "8013"]
```
azangru, line 9: Is this commented out line intentional? I can see that you are now passing this command in docker-compose.yaml; does this mean that gitlab-ci yaml will also need updating? Also, if you are now running this app in development with docker-compose, does the README need updating as well (it cites the docker command now)?

requirements.txt
```diff
@@ -3,3 +3,5 @@ pydantic>=2.4,<2.5
 uvicorn>=0.24,<0.25
```
azangru, line 3: Out of curiosity, is it worth updating the already listed dependencies (uvicorn is now at version 0.30 apparently)?

app/core/config.py
```diff
@@ -0,0 +1,56 @@
...
+
+from .logging import InterceptHandler
+import json
+
+VERSION = "0.0.0"
+API_PREFIX = "/api"
+
+config = Config(".env")
+DEBUG: bool = config("DEBUG", cast=bool, default=False)
+PROJECT_NAME: str = config("PROJECT_NAME", default="Ensembl Web Metadata API")
```
kamaldodiya: Should be Tools API

requirements.txt
```diff
@@ -3,3 +3,5 @@ pydantic>=2.4,<2.5
...
 aiohttp>=3.8,<3.9
 vcfpy>=0.13.8
```
kamaldodiya: Variation team uses specific tag on GitHub. @jonk-ebi Was suppose to make this change in the earlier PR.

kamaldodiya (commented):
Looks good. Does it need testing with review app ? or has it been tested locally ?

app/vep/vep_resources.py
```diff
@@ -0,0 +1,35 @@
...
+
+from core.logging import InterceptHandler
+
+logging.getLogger().handlers = [InterceptHandler()]
+
+router = APIRouter()
+
+
+
+@router.post("/submit", name="submit_vep")
```
azangru: Didn't we agree in the [spec](https://github.com/Ensembl/ensembl-web-tools-api/commit/eac68393828b4dedc7dc42f86c36f2e02eb2221a#diff-36c51008c7a160230713fa1fc74e66dd06b23a8e73fd42d4b742cbd119451b91R21) (see also the [Confluence doc](https://www.ebi.ac.uk/seqdb/confluence/display/ENSWEB/Decisions+for+new+website+APIs#DecisionsfornewwebsiteAPIs-Generaldecisions)) that the path will be `/submissions` (`/api/tools/vep/submissions`)?

requirements.txt
Mehrnaz-Charkhchi: I was tempted to do that @azangru but I couldn't make some of the BLAST features that are tied to uvicorn work and I made some adjustments to them , I'm not sure if a version update is gonna be a good idea. Andres should take a look at that later.

requirements.txt
Mehrnaz-Charkhchi: This PR doesn't have anything to do with vcfpy, We should keep an eye on it when reviewing Jyo's PR.

app/vep/vep_resources.py
Mehrnaz-Charkhchi: This is API is only for testing the file upload and submission API wasn't in the scope but I'll update the name.

Dockerfile
Mehrnaz-Charkhchi: Yes, it's intentional. I commented it exactly as reference in case we need to update the pipeline somewhere. 
Readme is updated now.

Mehrnaz-Charkhchi:
> Looks good. Does it need testing with review app ? or has it been tested locally ?

Maybe testing BLAST on a review branch would be a good idea, I did test it locally.

requirements.txt
kamaldodiya: Ah yes, true

kamaldodiya (approved):

app/blast/blast.py
```diff
@@ -0,0 +1,194 @@
...
+
+# @asynccontextmanager
+# async def lifespan(app: FastAPI):
+#     # Startup: setup data cache and requests session
+#     app.client_session = ClientSession()
+#     with open("/data/blast_config.json") as f:
+#         app.blast_config = json.load(f)
+#     yield
+#     # Shutdown: close requests session
+#     await app.client_session.close()
```
azangru, line 28: I would suggest deleting this commented out block. We can ask @veidenberg if he envisions any problems with migrating the client session from this function to the `run_blast` function.

PR #16, opened 2024-06-28
Submit a VEP workflow on Seqera platform

jyothishnt:
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2628


app/vep/utils/nextflow.py
```diff
@@ -0,0 +1,22 @@
+import requests
+import json
+from core.config import NF_TOKEN, NF_PIPELINE_URL, NF_COMPUTE_ENV_ID, NF_WORK_DIR
+from models.pipeline_model import PipelineParams
+
+def launch_workflow(pipeline_params: PipelineParams):
+    url = "https://api.cloud.seqera.io/workflow/launch" # ENV
```
Mehrnaz-Charkhchi: Please move url to configs

app/core/config.py
```diff
@@ -53,4 +53,10 @@
 
 logger.configure(handlers=[{"sink": sys.stderr, "level": LOGGING_LEVEL}])
 
-logging.info(BLAST_CONFIG)
\ No newline at end of file
+logging.info(BLAST_CONFIG)
```
Mehrnaz-Charkhchi, line 56: Please pull the latest updates from add_vep_app branch

app/vep/utils/nextflow.py
```diff
@@ -0,0 +1,21 @@
...
+    }
+
+    response = requests.post(SEQERA_WORKFLOW_LAUNCH_URL, headers=headers, json=pipeline_params)
+
+    # Check if successful
+    if response.status_code == 200:
+        return response.json()
+    else:
+        print(f"Failed to launch workflow. Status code: {response.status_code}")
+        return response.text
```
azangru: Shouldn't the function raise some kind of error if the request to the platform failed? If I am reading the code correctly, all errors will be quietly swallowed.

jyothishnt: As I said earlier, error handling is not required in this ticket, but since you mentioned I will add it.

azangru: > As I said earlier, error handling is not required in this ticket, but since you mentioned I will add it.

Sorry, after seeing the new commit, I realised that there must have been a bit of a misunderstanding. I thought the `launch_workflow` function would itself be called by some other function (the actual request handler); and the only responsibility of `launch_workflow` would be to raise an exception in case of a failed request to Seqera, which would be ultimately handled up the chain, in the request handler. I didn't expect that the `launch_workflow` function should actually return a Response object. I may well be wrong.

jyothishnt: hmm it makes sense for the launch_workflow to just raise an error and the caller handles it accordingly. Lets discuss in standup

Mehrnaz-Charkhchi (approved):

azangru (approved):

PR #18, opened 2024-07-03
Fix routing for VEP

kamaldodiya:
### Description
With the addition of the VEP it is now have 2 tools to handle blast/vep. 
This introduced the inconsistency in the way the route to /blast and /vep were being handled in production and development. 
This PR makes it consistent across environment and tools.

### Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2644

### Review App URL(s) 
This can be checked on dev-2020.ensembl.org

### Checklist

- [ ] Black formatting
- [x] Tests

### Dependencies 
The ingress rules now needs to be updated by removing the rewrite.

### DIFF
--- a/app/tests/test_blast.py
+++ b/app/tests/test_blast.py
@@ -3,13 +3,14 @@ import json
 import pytest
 from app.blast.blast import get_db_path
 from app.main import app
+from core.config import API_PREFIX
 
 # Test config endpoint
 def test_read_config():
 	with TestClient(app) as client: #include @startup hook
 		with open('/data/blast_config.json') as f:
 			config = json.load(f)
-		response = client.get('/api/tools/blast/config')
+		response = client.get(API_PREFIX + '/blast/config')
 		assert response.status_code == 200
 		assert response.json() == config
 
@@ -30,7 +31,7 @@ def test_get_db_path(blast_payload):
 # Test single BLAST job submission with a valid payload
 def test_blast_job(blast_payload):
 	with TestClient(app) as client:
-		response = client.post('/api/tools/blast/job', json=blast_payload)
+		response = client.post(API_PREFIX + '/blast/job', json=blast_payload)
 		assert response.status_code == 200
 		resp = response.json()
 		assert 'submission_id' in resp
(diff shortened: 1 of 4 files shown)


Mehrnaz-Charkhchi (approved):

app/tests/test_blast.py
```diff
@@ -9,7 +9,7 @@ def test_read_config():
 	with TestClient(app) as client: #include @startup hook
 		with open('/data/blast_config.json') as f:
 			config = json.load(f)
-		response = client.get('/blast/config')
+		response = client.get('/api/tools/blast/config')
```
veidenberg: I think we should use `API_PREFIX` everywhere instead of hardcoding.

veidenberg (changes requested):

veidenberg:
Will it break `/api/tools/vep/submissions` endpoint once ingress is updated since [it uses a different URL prefix](https://github.com/Ensembl/ensembl-web-tools-api/blob/24c9d074160e2b1d08dffdcb154238a53223640f/app/vep/routes.py#L24)?

app/tests/test_blast.py
kamaldodiya: @veidenberg  Done

kamaldodiya:
> Will it break `/api/tools/vep/submissions` endpoint once ingress is updated since [it uses a different URL prefix](https://github.com/Ensembl/ensembl-web-tools-api/blob/24c9d074160e2b1d08dffdcb154238a53223640f/app/vep/routes.py#L24)?

No it won't break. I have already update ingress on dev-2020 and manifests. @veidenberg

veidenberg:
> I have already update ingress on dev-2020 and manifests

Sorry I meant the [linked code line](https://github.com/Ensembl/ensembl-web-tools-api/blob/24c9d074160e2b1d08dffdcb154238a53223640f/app/vep/routes.py#L24) should also use the same `API_PREFIX`.

kamaldodiya:
> > I have already update ingress on dev-2020 and manifests
> 
> Sorry I meant the [linked code line](https://github.com/Ensembl/ensembl-web-tools-api/blob/24c9d074160e2b1d08dffdcb154238a53223640f/app/vep/routes.py#L24) probably also needs the same `API_PREFIX`.

No this doesn't need updating. This is app specific blast app will be at `/blast` and VEP will be at `/vep` 
The API will be at /api/<whatever> which is configurable this is global setting, in our case it is  `/api/tools`

veidenberg (approved):

PR #19, opened 2024-07-04
Create vep config ini file

jyothishnt:
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2627

This PR creates an ini file with vep configs in a temp location.


app/vep/models/pipeline_model.py
```diff
@@ -17,5 +20,19 @@ class LaunchParams(BaseModel):
     labelIds: List[str]
     headJobCpus: int = 1
     headJobMemoryMb: int = 4
+
 class PipelineParams(BaseModel):
     launch: LaunchParams
+
+class ConfigIniParams(BaseModel):
+  cache: int = 1
+  dir_cache: str = "/nfs/production/flicek/ensembl/variation/data/VEP/tabixconverted/"
```
kamaldodiya, line 36: Instead of hardcoding this can also be exposed through env variable.

azangru: > Instead of hardcoding this can also be exposed through env variable.

This is not something static and permanent though; it will eventually be a dynamic parameter that comes from the metadata api. Why put it in an env variable?

app/vep/vep_resources.py
```diff
@@ -15,21 +15,54 @@
...
-
 @router.post("/submissions", name="submit_vep")
 def submit_vep(genome_id:Annotated[str, Form()], input_file: UploadFile):
     try:
         return {"genome_id":genome_id,"filename": input_file.filename}
     except Exception as e:
         logging.info(e)
+
+# Creates config ini file
+def create_config_ini_file(symbol: bool = "true", biotype: bool = "true"):
```
azangru: Forgive the stupid question; but isn't `"true"` a string (`str`) rather than a `bool`? And don't we expect the actual booleans (e.g.`True`) as default?

By the way; I think the defaults should probably be false.

app/vep/models/pipeline_model.py
jyothishnt: yeah it should come from metadata api which is why I didnt bother to change.

app/vep/vep_resources.py
jyothishnt: yes, my bad. I didnt know it was `True` and not `true` which it didnt like.

app/core/config.py
```diff
@@ -61,3 +61,4 @@
 NF_PIPELINE_URL = config("NF_PIPELINE_URL")
 NF_WORK_DIR = config("NF_WORK_DIR")
 SEQERA_WORKFLOW_LAUNCH_URL = config("SEQERA_WORKFLOW_LAUNCH_URL")
+VEP_CONFIG_INI_PATH = config("VEP_CONFIG_INI_PATH")
```
kamaldodiya, line 64: This should be exposed to all the environment via tools-api.secret  in newwebsite-config repo. It doesn't have side affect so I will add it.

kamaldodiya (approved):

app/vep/vep_resources.py
azangru: That your editor allowed you to write `pullLatest: bool = "true"` without complaint probably suggests that type checking isn't working in your editor. I wonder if it's worth running type checking automatically upon commit as part of automated tests; or perhaps even as a pre-commit hook @kamaldodiya @Mehrnaz-Charkhchi @veidenberg what do you think?

app/core/config.py
kamaldodiya: This is done

app/vep/vep_resources.py
Mehrnaz-Charkhchi: Isn't a model the right place for this operation?

app/vep/models/pipeline_model.py
```diff
@@ -1,21 +1,65 @@
...
+  offline: int = 1
+  force_overwrite: int = 1
+  symbol: int = 1
+  biotype: int = 1
+  transcript_version: int = 1
+  canonical: int = 1
+  ini_file: str
+
+  # Creates config ini file
+  def create_config_ini_file(self, symbol: bool = False, biotype: bool = False):
```
azangru: How is this method going to be called?

If the caller is going to run:

```
config_ini = ConfigIniParams(symbol=True, biotype=True)
ini_file = config_ini.create_config_ini_file()
```

then the values of symbol and biotype will be passed into the class instance during its creation; and there will not be any need to pass any arguments into the `create_config_ini_file` method, which would read the values of symbol and biotype off the class instance fields.

(This also raises the question that probably the types of the symbol and biotype fields should be booleans.)

jyothishnt: At the moment that method doesn't return anything, instead it populates the class member variable `ini_file`
Regarding your suggestion @azangru I am not sure how ConfigIniParams work without BaseModel. @Mehrnaz-Charkhchi do you know?

Mehrnaz-Charkhchi (approved):

app/vep/vep_resources.py
Mehrnaz-Charkhchi: Didn't we agree on moving this to a model?

app/vep/models/pipeline_model.py
```diff
@@ -1,21 +1,70 @@
...
+  force_overwrite: int = 1
+  symbol: bool = False
+  biotype: bool = False
+  transcript_version: int = 1
+  canonical: int = 1
+  ini_file: str = None
+
+  @validator('symbol','biotype', pre = True, always = True)
+  def convert_to_int(cls, value: bool) -> int:
+    return 1 if (value) else 0
```
kamaldodiya: This has no effect as such 
https://docs.pydantic.dev/2.0/usage/types/booleans/

kamaldodiya (approved):

app/vep/models/pipeline_model.py
jyothishnt: Yeah true. Will remove it.

app/vep/models/pipeline_model.py
```diff
@@ -1,21 +1,66 @@
...
+biotype {biotype}
+transcript_version {self.transcript_version}
+canonical {self.canonical}
+'''
+    ini_file = tempfile.NamedTemporaryFile(prefix="vep_", suffix=".ini", dir=VEP_CONFIG_INI_PATH, delete=False)
+    try:
+      ini_file.write(config_yaml.encode())
+    finally:
+      ini_file.close()
+    self.ini_file = ini_file
```
azangru: The ini file arguably should be a value returned from the method rather than a class property. (Not that this should block the PR; just a thought).

azangru (approved):

app/vep/models/pipeline_model.py
```diff
@@ -1,21 +1,72 @@
...
+canonical {self.canonical}
+'''
+
+    try:
+      with tempfile.NamedTemporaryFile(prefix="vep_", suffix=".ini", dir=VEP_CONFIG_INI_PATH, delete=False) as ini_file:
+        ini_file.write(config_yaml.encode())
+      return ini_file
+    except Exception as e:
+      logging.info(e)
+      raise RuntimeError("Could not create vep config ini file") from e
```
veidenberg: I think `from e` is not needed here. It only [changes the wording of a sentence](https://stackoverflow.com/questions/24752395/python-raise-from-usage) in the printed error message (if we decide to print it to console).

jyothishnt: I am not sure I understood fully what is written in stackoverflow. But I trust you as you are far superior than me in python :-)

veidenberg (approved):

PR #20, opened 2024-07-05
Upload file to /submission endpoint

Mehrnaz-Charkhchi:
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2642


app/vep/models/upload_vcf_files.py
```diff
@@ -0,0 +1,49 @@
...
+    # Check file size
+    input_file.file.seek(0, 2)
+    file_size = input_file.file.tell()
+    input_file.file.seek(0)
+
+    if file_size > max_size:
+        raise HTTPException(
+            status_code=status.HTTP_413_REQUEST_ENTITY_TOO_LARGE,
+            detail=f"File size exceeds the maximum limit of {(max_size / (1024 * 1024)) + 10} MB."
+        )
```
azangru: Does the app need to be concerned with file size validation? Shouldn't it be the responsibility of ingress to set the maximum size of the request?

app/vep/vep_resources.py
```diff
@@ -11,25 +11,29 @@
 Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+<<<<<<< HEAD
```
azangru: ❓

app/vep/models/upload_vcf_files.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+async def validate_file(input_file: UploadFile = File(...), allowed_types: typing.Tuple[str] = ALLOWED_FILE_TYPES,
+                        max_size: int = MAX_FILE_SIZE) -> bool:
+    # Check file type
+    if not input_file.filename.endswith(ALLOWED_FILE_TYPES):
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=f"Invalid file type: {input_file.content_type}. Allowed types are: {', '.join(allowed_types)}."
+        )
```
azangru: The VEP tool should do input format checks on its own, regardless of file extension (see Nakib's comment on [Slack](https://genomes-ebi.slack.com/archives/C032E9YJ2N9/p1720433684433359)). I don't think we should reject files based on file extensions on our side.

app/vep/models/upload_vcf_files.py
```diff
@@ -0,0 +1,82 @@
...
+    def __call__(self, chunk: bytes):
+        self.body_len += len(chunk)
+        if self.body_len > self.max_size:
+            raise MaxBodySizeException(body_len=self.body_len)
+
+
+class Streamer:
+    def __init__(self, request):
+        self.request = request
+        self.filename = self.request.headers.get('Filename', None)
```
azangru: Is this an entirely custom header name, or did you see it used somewhere?

It seems that the standard header for this sort of thing is Content-Disposition. See the [article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition#as_a_header_for_a_multipart_body) on MDN:

image

The way that the article is phrased convinced me that you were right asking for the headers.

veidenberg: When looking at the docs and the form request payload, it turns out `Content-Disposition` headers with the original filename are already included by the browser with each POST request, but these headers are part of the form body itself. So using this information would need addition of the file renaming in server side with this line of code:
`os.rename(self.filepath, os.path.join(self.temp_dir, self._input_file.multipart_filename))` 
(file target instances have `.multipart_filename` property that is populated automatically from `content-disposition`)
It would mean less code both in client/server side and not reinventing the wheel, but we discussed that the issue with renaming is that it takes (some amount of) time? Sorry to come back to this but do we know if/how much time the renaming would actually take? Surely it must be faster than most other file actions we do (creating the file or streaming any amount of bytes to it)?

kamaldodiya: @veidenberg 

> Sorry to come back to this but do we know if/how much time the renaming would actually take? 

it takes around 380ms to rename 221MB file

```
$ ls -lh original_file 
-rw-r--r-- 1 6630 users 231M Mar 22  2023 original_file

$ cat rename_duration.py 
import os
src="/usr/data/original_file"
dst="/usr/data/renamed_file"
os.rename(src,dst)

$ time python rename_duration.py 
real	0m0.038s
user	0m0.012s
sys	0m0.006s

$ ls -lh renamed_file 
-rw-r--r-- 1 6630 users 231M Mar 22  2023 renamed_file
```

azangru: > it takes around 380ms to rename 221MB file
```
real	0m0.038s
user	0m0.012s
sys	0m0.006s
```
380 milliseconds is a lot. But isn't `0m0.038s` 38 milliseconds rather than 380?

Also, I ssh-ed into a pod, and timed python execution of a dummy script that just imports os and immediately exits, to account for a cold startup time:

```
time python -c 'import os'

real    0m0.017s
user    0m0.014s
sys     0m0.003s
```

Does it mean that the actual renaming of a file takes 38 - 17, or roughly 20 milliseconds?

veidenberg: In my computer (average of 1000 repeats):
Renaming a 1-byte file: 0.68 milliseconds
Renaming a 500MB file: 0.66 ms (changing the file type)
Creating a 1-byte file: 1.3 ms

kamaldodiya: @azangru  Yes you are right its 38ms. That was at last night 9pm I was a running that script probably in sleepy mode ;)

azangru: > In my computer (average of 1000 repeats):

We should probably consider containers, nfs, and such. But I agree with you that it shouldn't be a time-consuming operation.

veidenberg: Yes, the numbers will have a multiplier in k8s but it indicates that file renaming is one of the fastest file operations one can do (and the file size has no effect) so it should to be a non-issue for us. Which, I think, leaves this as our new best option for handling the temp file name (instead of using custom headers).

kamaldodiya: @veidenberg I ran that script in k8s with NFS so it is as realistic as it can get.

veidenberg: I know @kamaldodiya, I was just lazy :) The same 1000-cycle test in tools-api k8s pod with NFS (using perf_counter module inside the script for accuracy):
- Rename small file: 1.06ms
- Rename large file: 1.16ms
- Create small file: 2.33ms

The point stands: we should use the file renaming approach.

app/vep/models/upload_vcf_files.py
```diff
@@ -0,0 +1,87 @@
...
+        body_validator =MaxBodySizeValidator(MAX_REQUEST_BODY_SIZE)
+        try:
+            self.parser.register('input_file', self._input_file)
+            self.parser.register('parameters', self.parameters)
+            self.parser.register('genome_id', self.genome_id)
+
+            async for chunk in self.request.stream():
+                body_validator(chunk)
+                self.parser.data_received(chunk)
+            return True
```
veidenberg, line 84:
```suggestion
           
            os.rename(self.filepath, os.path.join(self.temp_dir, self._input_file.multipart_filename))
            return True
```
@Mehrnaz-Charkhchi this should work (see the discussion above).

app/vep/models/upload_vcf_files.py
Mehrnaz-Charkhchi: @veidenberg , I don't understand what is wrong with using custom headers? 
We already have file name after we receive the first chunk of the body and we have discussed this in code-review, What is the benefit of adding this completely unnecessary operation which can cause issues in file type, data , ... ?

app/vep/models/upload_vcf_files.py
Mehrnaz-Charkhchi: If I'm already using header Filename what is the point of renaming?

app/vep/models/upload_vcf_files.py
azangru: @Mehrnaz-Charkhchi I would ask the same questions but turn them the other way around:
- What's the point of adding a custom header when file name information already exists in the payload?
- Why add extra complexity to the payload and to the client-side code?
- Why add a non-standard header to the request?
- Why should the renaming of a file cause issues for file type, data, etc.?

app/vep/models/upload_vcf_files.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class Streamer:
+    def __init__(self, request):
+        self.request = request
+        self.filename = self.request.headers.get('Filename', None)
+        self.file_name_validator(self.filename)
+        self.temp_dir = tempfile.mkdtemp(dir=VCF_UPLOAD_DIRECTORY)
+        self.filepath = os.path.join(str(self.temp_dir), os.path.basename(self.filename))
+        self._input_file = FileTarget(self.filepath)
```
veidenberg, line 58: I think leading underscore is not needed here as `_input_file` is not a private variable - it will be sent to Nextflow pipeline (via ini conf file).

app/vep/models/upload_vcf_files.py
veidenberg: This is the code line for the approach where we don't implement custom headers (see the discussion above). We could use this as a fallback (rename only when no filename in the header) to test both approaches (header approach/renaming approach)

app/vep/models/upload_vcf_files.py
Mehrnaz-Charkhchi: For the streaming operation this specific variable doesn't need to be accessible outside of the class since it's only used to stream the data to the file. We have filepath available for that.

app/vep/models/upload_vcf_files.py
```diff
@@ -75,6 +76,11 @@ async def stream(self):
             async for chunk in self.request.stream():
                 body_validator(chunk)
                 self.parser.data_received(chunk)
+
+            if self.filename == 'temp_name':
+                os.rename(self.filepath, os.path.join(self.temp_dir, self._input_file.multipart_filename))
+                self.filename = self._input_file.multipart_filename
+                self.filepath = os.path.join(self.temp_dir, self._input_file.multipart_filename)
```
azangru: these three lines are repeating the long and non-obvious incantation `self._input_file.multipart_filename` three times, and another incantation `os.path.join(self.temp_dir, self._input_file.multipart_filename))` two times. Don't you want to assign them to easily readable variables and then use those?

```
filename = self._input_file.multipart_filename # read the file name from the multipart form data
updated_file_path = os.path.join(self.temp_dir, filename)

os.rename(self.filepath, updated_file_path)
self.filename = filename
self.filepath = updated_file_path
```

app/vep/models/upload_vcf_files.py
veidenberg: Ah yes.

kamaldodiya (approved):
Looks good to me. Just one thing is that if test_obj  is the appropriate name in /submissions. Rename if you could or else its good

veidenberg (approved):
Looks good to me.

app/vep/models/upload_vcf_files.py
Mehrnaz-Charkhchi: I'm trying to avoid unnecessary variables that will only serve as a temp data holder. 
But self.filepath can be simplified.
`self.filepath = os.path.join(self.temp_dir, self.filename)`

azangru (approved):

PR #22, opened 2024-07-12
Submission integration

kamaldodiya:
### Description
This PR integrates different modules needed for submissions endpoint
- Generating .ini file
- Streaming user uploaded data
- Create a workflow in the Seqera Platform
- Returns Workflow ID

### Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2667

### Review App URL(s) 
None

### Example(s)

```
$ curl 'http://localhost:8013/api/tools/vep/submissions' \
-X 'POST' \     
-H 'Content-Type: multipart/form-data; boundary=----WebKitFormBoundary2RyyLdAFakK7ucEc' \
-H 'Accept: */*' \
--data-binary $'------WebKitFormBoundary2RyyLdAFakK7ucEc\r\nContent-Disposition: form-data; name="genome_id"\r\n\r\na7335667-93e7-11ec-a39d-005056b38ce3\r\n------WebKitFormBoundary2RyyLdAFakK7ucEc\r\nContent-Disposition: form-data; name="input_file"; filename="Venter_chr1.vcf"\r\nContent-Type: text/vcard\r\n\r\n\r\n------WebKitFormBoundary2RyyLdAFakK7ucEc\r\nContent-Disposition: form-data; name="parameters"\r\n\r\n{"transcript_set":"gencode_comprehensive","symbol":false,"biotype":false}\r\n------WebKitFormBoundary2RyyLdAFakK7ucEc--\r\n'  | jq
{
  "submission_id": "2oMMWmZO9pRpZH"
}

# Uploaded files
$ ls -l tmp16ozskvl/
-rw-r--r--  1 kamal  staff    0 12 Jul 12:27 Venter_chr1.vcf
-rw-r--r--  1 kamal  staff  219 12 Jul 12:27 config.ini

# config.ini
$ cat tmp16ozskvl/config.ini 
cache 1
dir_cache /nfs/production/flicek/ensembl/variation/data/VEP/tabixconverted/
species homo_sapiens
assembly GRCh38
cache_version 110
offline 1
force_overwrite 1
symbol 0
biotype 0
transcript_version 1
canonical 1
```

### Checklist

- [x] Black formatting
- [x] Tests


app/vep/models/pipeline_model.py
```diff
@@ -63,7 +63,7 @@ def create_config_ini_file(self):
 '''
 
     try:
-      with tempfile.NamedTemporaryFile(prefix="vep_", suffix=".ini", dir=UPLOAD_DIRECTORY, delete=False) as ini_file:
+      with tempfile.NamedTemporaryFile(prefix="vep_", suffix=".ini", dir=dir, delete=False) as ini_file:
```
Mehrnaz-Charkhchi: Do we need a NamedTemporaryFile anymore or can we go with a simple  config.ini file in temp dir?

app/vep/vep_resources.py
```diff
@@ -34,8 +36,15 @@ async def submit_vep(request : Request):
...
+        genome_id = stream_obj.genome_id.value.decode()
+
+        parameters_dict = json.loads(parameters)
+        ini_parameters = ConfigIniParams(**parameters_dict)
+        inifile = ini_parameters.create_config_ini_file(stream_obj.temp_dir)
+        vep_config_parameters = VEPConfigParams(vcf=stream_obj.filepath,vep_config=inifile.name)
+        launch_params = LaunchParams(paramsText=vep_config_parameters, labelIds=[])
+        pipeline_params = PipelineParams(launch=launch_params)
         if stream_result:
             return {"message": f"Successfully uploaded{stream_obj.filepath}"}
```
veidenberg: According to APISpec this should return [JSON with submission ID](https://github.com/Ensembl/ensembl-web-tools-api/blob/cbfea2ef333a491ba9aa66ecece27cc36fdaf889/APISpecification.yaml#L280). We can use a dummy ID until proper workflow submission is implemented in [ENSWBSITES-2656](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2656) (in that case update the PR description).

app/vep/models/pipeline_model.py
```diff
@@ -12,10 +12,17 @@ class VEPConfigParams(BaseModel):
   vcf: str
   vep_config: str
 
+  @model_serializer
+  def vep_config_serialiser(self):
```
veidenberg: This is not needed. You can use [model_dump_json()](https://docs.pydantic.dev/latest/concepts/serialization/#modelmodel_dump_json) instead.

app/vep/models/pipeline_model.py
veidenberg: I'd also prefer simple `config.ini` here.

app/vep/utils/nextflow.py
```diff
@@ -12,9 +12,9 @@ def launch_workflow(pipeline_params: PipelineParams):
         "Content-Type": "application/json",
         "Authorization": f"Bearer {NF_TOKEN}"
     }
-
+    payload = pipeline_params.model_dump()
```
veidenberg, line 15: `pipeline_params.model_dump_json()`

app/vep/models/pipeline_model.py
kamaldodiya: @veidenberg  The seqera platform expects stringify encoded json. I tried all combination but couldn't generated it so I have to write that (ugly) code. :(

veidenberg: What about `json.dumps(pipeline_params.model_dump_json())`?

kamaldodiya: This will lead to infinite recursion. I tried all different combination but this is what finally working :(

veidenberg: No it won't, it will just stringify twice, escaping the `"` chars, if that's what nextflow expects.

app/vep/models/pipeline_model.py
```diff
@@ -1,55 +1,69 @@
...
-  vcf: str
-  vep_config: str
+    vcf: str
+    vep_config: str
+
+    @model_serializer
+    def vep_config_serialiser(self):
+        vcf_str = '"vcf":"' + self.vcf + '"'
+        config_str = '"vep_config":"' + self.vep_config + '"'
+        stringified_encoded_json = "{" + vcf_str + "," + config_str + "}"
```
azangru, line 23: Is this the preferred way of concatenating/interpolating/templating strings? You don't prefer f-strings?

app/vep/models/pipeline_model.py
```diff
@@ -1,55 +1,69 @@
...
+    offline: int = 1
+    force_overwrite: int = 1
+    symbol: bool = False
+    biotype: bool = False
+    transcript_version: int = 1
+    canonical: int = 1
 
-  # Creates config ini file
-  def create_config_ini_file(self):
+    # Creates config ini file
```
azangru: Doesn't the method name already tell what the method does? What's the value of this comment?

app/core/config.py
```diff
@@ -29,14 +29,14 @@
 API_PREFIX = "/api/tools"
 
 config = Config(".env")
-DEBUG: bool = config("DEBUG", cast=bool, default=False)
+DEBUG: bool = config("DEBUG", cast=bool, default=True)
```
veidenberg: Why in debug mode when not specified?

app/vep/models/pipeline_model.py
```diff
@@ -60,12 +74,12 @@ def create_config_ini_file(self):
...
 
-    try:
-      with tempfile.NamedTemporaryFile(prefix="vep_", suffix=".ini", dir=UPLOAD_DIRECTORY, delete=False) as ini_file:
-        ini_file.write(config_yaml.encode())
-      return ini_file
-    except Exception as e:
-      logging.info(e)
-      raise RuntimeError("Could not create vep config ini file")
+        try:
+            with open(os.path.join(dir, "config.ini"), "w") as ini_file:
```
azangru, line 79: Out of curiosity and ignorance: is this any better than the tempfile approach?

app/report.xml
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="utf-8"?><testsuites><testsuite name="pytest" errors="0" failures="1" skipped="0" tests="10" time="88.308" timestamp="2024-07-11T14:04:44.670455" hostname="C02DP9MZML7H"><testcase classname="tests.test_blast" name="test_read_config" time="0.025" /><testcase classname="tests.test_blast" name="test_get_db_path" time="0.002" /><testcase classname="tests.test_blast" name="test_blast_job" time="7.081" /><testcase classname="tests.test_blast" name="test_blast_jobs" time="4.009" /><testcase classname="tests.test_blast" name="test_blast_job_data_error" time="0.008" /><testcase classname="tests.test_blast" name="test_blast_job_jd_error" time="0.169" /><testcase classname="tests.test_blast" name="test_blast_job_status" time="75.012"><failure message="assert 500 == 200&#10; +  where 500 = &lt;Response [500 Internal Server Error]&gt;.status_code">def test_blast_job_status():
```
veidenberg: Add to gitignore?

app/vep/vep_resources.py
```diff
@@ -34,10 +36,18 @@ async def submit_vep(request : Request):
...
+        parameters = stream_obj.parameters.value.decode()
+        genome_id = stream_obj.genome_id.value.decode()
+
+        parameters_dict = json.loads(parameters)
+        ini_parameters = ConfigIniParams(**parameters_dict)
+        inifile = ini_parameters.create_config_ini_file(stream_obj.temp_dir)
+        vep_config_parameters = VEPConfigParams(vcf=stream_obj.filepath,vep_config=inifile.name)
+        launch_params = LaunchParams(paramsText=vep_config_parameters, labelIds=[])
+        pipeline_params = PipelineParams(launch=launch_params)
         if stream_result:
```
Mehrnaz-Charkhchi, line 51: Should we add some checks for ini_parameters.create_config_ini_file() 's result here?

app/vep/models/pipeline_model.py
veidenberg: Tempfile would just add complexity with no benefit over static filename (random filename, unneeded operations run by tempfile). We are already using tempdir.

app/tests/test_blast.py
```diff
@@ -5,116 +5,129 @@
...
-		assert 'submission_id' in resp
-		assert 'jobs' in resp
-		assert len(resp['jobs']) == 1
-		job = resp['jobs'][0]
-		assert 'sequence_id' in job
-		assert 'genome_id' in job
-		assert 'job_id' in job
-		assert job['job_id'].startswith('ncbiblast')
-		assert 'sequence_id' in job and job['sequence_id'] == 1
+    with TestClient(app) as client:
```
veidenberg: Lots of whitespace noise by Black formatting. Can you do the the formatting in a separate PR (or a commit outside this PR)?

app/vep/vep_resources.py
```diff
@@ -23,29 +23,44 @@
...
 logging.getLogger().handlers = [InterceptHandler()]
 
 router = APIRouter()
 
 
 @router.post("/submissions", name="submit_vep")
-async def submit_vep(request : Request):
+async def submit_vep(request: Request):
     try:
         stream_obj = Streamer(request=request)
```
azangru: Nitpicking: I would rename, or question, some things here:

- `stream_obj` -> `streamer` (same name as class, but lower-cased), or `request_streamer`. Otherwise the name is misleading. It is a streamer itself, not a stream.
- `stream_result` -> `has_stream_completed` (to make it obvious that the return value is a boolean). Although I would drop this variable altogether. Looking at the implementation of the stream method, it can only ever return `True`, or it can error out. But you are already handling your exceptions. What's the point in checking this return value as well?
- `parameters` -> `vep_parameters`, or `vep_job_parameters` (there are so many parameters in this function, ugh)

app/vep/models/pipeline_model.py
kamaldodiya: I think initial assumption was to store in a UPLOAD_DIRECTORY but it make sense to have it in the same directory as vcf data since the directory is unique.

app/vep/vep_resources.py
kamaldodiya: What sort of check @Mehrnaz-Charkhchi  ?

app/tests/test_blast.py
kamaldodiya: That is the price I have to pay since this is the first time I did black formatting.

app/vep/models/pipeline_model.py
kamaldodiya: I did try all combination @veidenberg  but it's not working.  The only thing which is working is what we have in the code right now. It just needs paramsText as stringified_encoded_json ( with escaped ") hard to achieve with any combination of json, dumps, loads, model_dump, model_dump_json. 
I tried one more time and now getting 400 as the payload is different. 
```
"{\"launch\":{\"computeEnvId\":\"17wuWKrWOZW5JrLoGGwPdD\",\"pipeline\":\"https://github.com/jonk-ebi/platform-test-pipeline\",\"workDir\":\"/homes/ens_adm02/runs\",\"revision\":\"main\",\"pullLatest\":true,\"configProfiles\":[\"slurmnew\"],\"paramsText\":{\"vcf\":\"/Users/kamal/Development/e2020/cicd/e2s/ensembl-web-tools-api/tmpdy1zmjo9/Venter_chr1.vcf\",\"vep_config\":\"/Users/kamal/Development/e2020/cicd/e2s/ensembl-web-tools-api/tmpdy1zmjo9/config.ini\"},\"preRunScript\":\"module load nextflow\",\"postRunScript\":\"\",\"stubRun\":false,\"labelIds\":[],\"headJobCpus\":1,\"headJobMemoryMb\":4}}"
```

Error

```
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://api.cloud.seqera.io/workflow/launch
```

app/core/config.py
```diff
@@ -36,7 +36,7 @@
     cast=CommaSeparatedStrings,
     default="*",
 )
-with open("/data/blast_config.json") as f:
+with open("../data/blast_config.json") as f:
```
jyothishnt: is this correct?

app/tests/test_blast.py
veidenberg: That's the point. You don't have to pay the price in this PR and instead commit formatting after PR is merged (or commit to dev branch and rebase). Most of the changes here are now unrelated to this PR. Not very important of course, just the amount of noise makes relevant changes harder to follow (especially in the future).

app/core/config.py
azangru: It'll look from inside the app directory into the data directory that's sitting next to it in the repo; should be correct 

image

app/vep/vep_resources.py
```diff
@@ -23,29 +23,44 @@
...
+        )
     except Exception as e:
         print(e)
-        return HTTPException(status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
-                            detail='There was an error uploading the file.')
-
-
+        return HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="There was an error uploading the file.",
```
veidenberg: This message should be updated since it also catches other errors (e.g. thrown when creating ini file or submitting workflow).

jyothishnt (approved):
Looks ok to me. Can be merged once other comments are addressed

app/core/config.py
kamaldodiya: did that for local setup/testing shouldn't be this way. Corrected

app/vep/models/pipeline_model.py
kamaldodiya: With formatting the variable should be enclosed in { } but I need it in the string itself . Also, the way the Pedantic serialiser works this is the only thing which is working I am afraid. I will try to play around with formatting option and see if it works.

app/vep/vep_resources.py
kamaldodiya: Done

azangru (approved):

app/tests/test_blast.py
kamaldodiya: Well that's jut 2 files from blast. Rest all are rom VEP which very much is being worked on in this PR. Assume this is that PR :) If not then I will remove those two files fro this  PR. :)

veidenberg: It's much more that just blast files (see the [commit](https://github.com/Ensembl/ensembl-web-tools-api/commit/3f41f022dc454bcec800a30b96e9175be14076ae)). Moving formatting out of this PR was just a suggestion (for readability), if you wish to keep it here I won't object :)

kamaldodiya: @veidenberg  Removed blast  files it now so only changes related to VEP. :)

Mehrnaz-Charkhchi (approved):

app/tests/test_blast.py
veidenberg: Removing just the blast files is worst of both worlds: we still have the whitespace noise in this PR and yet have to create a separate commit/PR for whitespace formatting in blast files. Please put back the blast files if you wish to do the formatting in this PR. Thanks!

veidenberg (approved):

PR #23, opened 2024-07-16
Add spec for VEP submission status endpoint to the api spec

azangru:
## Description
Added specifications for VEP submission status endpoint to the OpenAPI spec.

### Knowledge base
See Seqera's [openAPI spec](https://cloud.seqera.io/openapi/index.html#get-/workflow/-workflowId-) or [code documentation](https://github.com/seqeralabs/tower-cli/blob/50dd04dd806acf8461dfc9b286faec7e874aec15/USAGE.md?plain=1#L254-L259) for the list of possible values that can come from Seqera. The strings reported by Seqera are all upper-cased.

### NOTE
Among the possible statuses that can be returned by Seqera is `UNKNOWN`. I have no idea when it may occur (docs don't explain anything other than that it is an "indeterminate status"), or what to do with it on the client side. I suggest that if tools api ever encounters it, it should consider the job as failed, and respond to the web client with the `FAILED` status.

### Example json response

```json
{
  "submission_id": "string",
  "status": "SUBMITTED"
}
```

## Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2648

### DIFF
--- a/APISpecification.yaml
+++ b/APISpecification.yaml
@@ -305,7 +305,7 @@ components:
       properties:
         message:
           type: string
-          description: A message that says, "A submission with id <id> was not found"
+          example: A submission with id 1234567 was not found
     VepSubmissionStatusResponse:
       type: object
       required:
--- a/APISpecification.yaml
+++ b/APISpecification.yaml
@@ -42,6 +42,21 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/VepFormSubmissionCreatedResponse'
+  /api/tools/vep/submissions/{id}/status:
+    get:
+      parameters:
+        - name: id
+          in: path
+          required: true
+          schema:
+            type: string
+      responses:
+        '200':
+          description: The request has succeeded.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/VepSubmissionStatusResponse'
   /api/tools/vep/submissions/{id}/results:
     get:
       description: Returns results of VEP analysis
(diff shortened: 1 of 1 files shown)


APISpecification.yaml
```diff
@@ -277,6 +292,27 @@ components:
...
+      required:
+        - submission_id
+        - status
+      properties:
+        submission_id:
+          type: string
+        status:
+          type: string
+          description: >
+            The possible statuses for a VEP submission are based on the list of documented statuses of Seqera workflows
```
veidenberg, line 320: This should be a short summary of this response. Implementation details/suggestions will go to the JIRA ticket that builds this enpoint.

APISpecification.yaml
```diff
@@ -277,6 +292,27 @@ components:
...
+            (see https://github.com/seqeralabs/tower-cli/blob/50dd04dd806acf8461dfc9b286faec7e874aec15/USAGE.md?plain=1#L254-L259),
+            except for the 'UNKNOWN' status, which is listed in Seqera docs without explanation how a workflow can get into that status.
+            If the tools api receives 'UNKNOWN' status from the Seqera platform, interpret it the same as FAILED.
+          enum:
+            - SUBMITTED
+            - RUNNING
+            - SUCCEEDED
+            - FAILED
+            - CANCELLED
     VepFormSubmissionCreatedResponse:
```
veidenberg, line 330: What was our descision about error responses? Do we include these in the spec from the start, not bother (assume that e.g. 404 and 500 are implemented for all endpoints and expected/handled by client) or wait for BE to add later (e.g. during implementation)?

veidenberg (approved):
Looks good, added some philosphical comments.

APISpecification.yaml
azangru: Hmm, I should probably add a 404 here, because it would be different from how Seqera will respond.

kamaldodiya (approved):

Mehrnaz-Charkhchi (approved):

PR #25, opened 2024-07-19
Add pipeline status endpoint

Mehrnaz-Charkhchi:
### Description
When the client asks, query the Seqera platform, and report the status of the workflow to the client.

### Related JIRA Issue(s)
[ENSWBSITES-2649](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2649)


app/core/config.py
```diff
@@ -60,5 +60,6 @@
 NF_PIPELINE_URL = config("NF_PIPELINE_URL")
 NF_WORK_DIR = config("NF_WORK_DIR")
 SEQERA_WORKFLOW_LAUNCH_URL = config("SEQERA_WORKFLOW_LAUNCH_URL")
+SEQERA_API = config("SEQERA_API")
```
kamaldodiya, line 62: There are two URLs for which base (HOST) will be same `SEQERA_WORKFLOW_LAUNCH_URL` and  `SEQERA_API`.
Would it be possible to just have `SEQERA_API` and remove `SEQERA_WORKFLOW_LAUNCH_URL` and define the LAUNCH_URL from `SEQERA_API`.
If not in this PR then may be in future

Mehrnaz-Charkhchi: I did think about it @kamaldodiya and I agree with you. I thought maybe after we get all the approvals for this PR I can make that update to reduce the noise.

app/vep/utils/nextflow.py
```diff
@@ -24,9 +24,28 @@ def launch_workflow(pipeline_params: PipelineParams):
...
+        _seqera_workflow_status_url = f"{SEQERA_API}/workflow/{job_id}"
+        response = requests.get(_seqera_workflow_status_url, headers=_headers)
+
+        response.raise_for_status()
+        pipeline_status = PipelineStatus(job_id=job_id, status=response.json())
+        return pipeline_status
+    except HTTPError as e:
+        raise e
+    except Exception as e:
+        logger.exception("VEP connection error: ", e)
```
kamaldodiya, line 50: If the purpose here is to log the connection issue while requesting seqera API then it would be good to be explicit and say connection issue with seqera API while logging

Mehrnaz-Charkhchi: It's a bit more general than that because we're not connecting to seqera here. It can be any unexpected issue that happened while connecting to this API exactly like line 31.

kamaldodiya (approved):

app/vep/vep_resources.py
```diff
@@ -54,12 +61,30 @@ async def submit_vep(request: Request):
...
+
+
+@router.get("/submissions/{job_id}/status", name="submission_status")
+async def submit_vep(request: Request, job_id: str):
+    try:
+        workflow_status = await get_workflow_status(job_id)
+        return JSONResponse(content=workflow_status.dict())
+
+    except HTTPError as http_error:
+        if http_error.response.status_code == 403:
```
azangru: Didn't we see that Seqera's response to a missing workflow id can also be a `400`? We would also want to transform that into a 404.

app/vep/vep_resources.py
```diff
@@ -54,12 +61,30 @@ async def submit_vep(request: Request):
...
+            response_msg = json.dumps(
+                {
+                    "status_code": status.HTTP_404_NOT_FOUND,
+                    "details": f"A submission with id {job_id} was not found",
+                }
```
azangru, line 81: This isn't what I wrote in the spec 🤷‍♂️ Are we changing the spec?

app/vep/vep_resources.py
```diff
@@ -54,12 +61,30 @@ async def submit_vep(request: Request):
...
-            detail="Could not submit VEP job/workflow.",
-        )
+        return response_error_handler(result={"status": 500})
+
+
+@router.get("/submissions/{job_id}/status", name="submission_status")
+async def submit_vep(request: Request, job_id: str):
+    try:
+        workflow_status = await get_workflow_status(job_id)
+        return JSONResponse(content=workflow_status.dict())
```
azangru, line 73: I may be misreading the python here; but what's the shape of the response going to be? I don't think it conforms to the spec.

app/vep/vep_resources.py
Mehrnaz-Charkhchi: In the spec for this specific error  we're drifting from the general error format that we use, I'm ready to either updating the spec or code but I thought you should know that before.

app/vep/vep_resources.py
Mehrnaz-Charkhchi: Done!

app/vep/vep_resources.py
Mehrnaz-Charkhchi: What do you mean?
This response is not in the spec?
`{
    "job_id": "5SYhtjFhDfmzx2",
    "status": "FAILED"
}`

azangru: The spec calls the field `submission_id` 😉

app/vep/vep_resources.py
azangru: I did not know about the general error format. Personally, I think the status_code field in the body is excessive (the code is already present in the header); but if that's our general error format... Let's see what others think.

app/vep/vep_resources.py
Mehrnaz-Charkhchi: Ah, I got confused by having two names for submission_id, in the url we call it job_id and in the response we call it submission_id! we should update one of them!

app/vep/vep_resources.py
Mehrnaz-Charkhchi: As far as I remember resolver, refget proxy and metadata API are using the same error_response file that generates the same format I thought you're aware of it!  Let's go ahead with using the general format for now then since others are already using it.

app/vep/vep_resources.py
Mehrnaz-Charkhchi: Done

app/vep/vep_resources.py
Mehrnaz-Charkhchi: spec is updated.

azangru (approved):

app/vep/vep_resources.py
```diff
@@ -54,12 +61,30 @@ async def submit_vep(request: Request):
...
+                }
+            )
+            return PlainTextResponse(
+                response_msg, status_code=status.HTTP_404_NOT_FOUND
+            )
+
+        return response_error_handler(result={"status": http_error.response.status_code})
+    except Exception as e:
+        logging.debug(e)
+        return response_error_handler(result={"status": 500})
```
veidenberg, line 90: `status.HTTP_500_INTERNAL_SERVER_ERROR` to make it consistent withg the other responses.

app/vep/vep_resources.py
```diff
@@ -54,12 +61,30 @@ async def submit_vep(request: Request):
...
-            status_code=status.HTTP_413_REQUEST_ENTITY_TOO_LARGE,
-            detail="Maximum file size limit exceeded.",
-        )
+        return response_error_handler(result={"status": 413})
     except Exception as e:
-        return HTTPException(
-            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
-            detail="Could not submit VEP job/workflow.",
-        )
+        return response_error_handler(result={"status": 500})
```
veidenberg, line 66: Are we using integer status codes or constants (`status.HTTP_500_INTERNAL_SERVER_ERROR`)?

app/vep/vep_resources.py
veidenberg: I was confused why we're mixing integer and constant status codes in this file, but it seems that it's because we are reimplementing `http_404_not_found` function from  `response_error_handler` just to change the value of the `details` field. Would it make sense to instead update `response_error_handler` to accept a custom status message?

jyothishnt (approved):

PR #26, opened 2024-07-26
Results download

kamaldodiya:
### Description
This PR

- Update status endpoint to use workspace ID to get the status of workflow
- Update PipelineStatus to return nextflow response as json
- /download endpoint to download file ( Not yet tested )
    -  check the status of the workflow and  if it is succeeded
         -  determine the name of the out file  and its path
- Refactor few variable names

### Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2658

### Review App URL(s) 
https://dev-2020.ensembl.org

### Checklist

- [x] Black formatting
- [x] Tests

### Dependencies 

This will need the workspace id to be exposed through environment variable.


app/vep/vep_resources.py
```diff
@@ -88,3 +88,40 @@ async def vep_status(request: Request, submission_id: str):
     except Exception as e:
         logging.debug(e)
         return response_error_handler(result={"status": 500})
+
+@router.get("/download/{submission_id}", name="download_results")
```
azangru: We haven't written a spec; but following the template of our other VEP endpoints, this should probably be `/submissions/:id/download` within the `/api/tools/vep` namespace

app/vep/vep_resources.py
```diff
@@ -88,3 +88,40 @@ async def vep_status(request: Request, submission_id: str):
...
+async def download_results(request: Request, submission_id: str):
+    try:
+        workflow_status_response = await get_workflow_status(submission_id)
+        workflow_status = workflow_status_response.dict()
+        # To use out file it will require changes in nextflow and get_workflow_status endpoint
+        results_file = workflow_status['outfile']
```
azangru: I understand that this is the result of a previous PR, and is out of scope in this PR; but reading the code, I realised that backend is using the word "status" here in two senses. One sense is the whole information object about a workflow — this is the return value of `workflow_status_response.dict()`. The other sense is the actual status of the submission, which you are accessing via `workflow_status['status']`. I personally would have found this confusing; and would have probably called the `workflow_status_response` something like `workflow_info_response` or `workflow_record_response`, similar to what it's called in the Seqera api (and would have made similar name changes in the model). I am leaving it entirely up to you to consider whether this ambiguity of meaning is something that is bothering you as well.

app/vep/vep_resources.py
```diff
@@ -88,3 +88,40 @@ async def vep_status(request: Request, submission_id: str):
...
+        if workflow_status['status'] == "SUCCEDED":
+            return await FileResponse(results_file)
+        else:
+            response_msg = json.dumps(
+                {
+                    "status_code": 200,
```
azangru: A 404 response with a status_code 200 in the body is evil. (This is why I don't understand the reason you include the status_code field in the body of error responses.)

veidenberg: This will only be sent as 200 response (on line 124). Errors are handled below.

azangru: When I left my comment, the return value was:

image

(see [commit d17c462](https://github.com/Ensembl/ensembl-web-tools-api/pull/26/commits/d17c4620a0429c0ec3efcfc28edea4e9104565ec#diff-d4096798e7cffcfe65307ad13fd12e1c101f522d30bd91a73fd58ffccae78086R108-R110))

_(I still don't understand why you guys are putting status codes in the response body though)_

azangru: Also, reading through the code again, I do not think the successful response from the endpoint should include two different types of data (a file or a json). I think Kamal's initial version with a 404 response was correct (the file does not exist).

kamaldodiya: > I still don't understand why you guys are putting status codes in the response body though

This now removed from the response body.
https://github.com/Ensembl/ensembl-web-tools-api/blob/457499e601193d28d71b9215c0396950f562ece1/app/vep/vep_resources.py#L85-L87
It is very much needed in the PainTextResonse quoted above as that is the http_status code response.

azangru: As the code currently stands, it can respond with a successful (200) response that can be either a file itself, or a plain-text string saying that the file is not yet ready. This design worries me. It means that the client should be capable of inspecting the response and making decisions based on whether it's a file or a plain-text string.

Could we please change the design of the endpoint such that if the file is not available, the response is a 404? The message in the 404 response can say that the analysis for the submission with provided id has not yet finished.

kamaldodiya: Fair point. This has been reflected now. I think download endpoint should return the file if the job is succeeded and else say 404 the result now found. The task to check the status of the workflow should lie with the /status endpoint and client should use that. otherwise download will become download + status both.
In this case client has requested a file if it exists then it should return it or else say file not found.

azangru (approved):

app/vep/vep_resources.py
```diff
@@ -70,21 +74,70 @@ async def submit_vep(request: Request):
...
+                media_type="application/gzip",
+                filename=results_file_path.name,
+            )
+        else:
+            response_msg = json.dumps(
+                {
+                    "details": f"A submission with id {submission_id} is not yet finished",
+                }
+            )
+            return PlainTextResponse(
```
Mehrnaz-Charkhchi: I think this should be a JSON response so it can produce a valid 404.

kamaldodiya: This is now changed for both /download and /status endpoint  to maintain consistency.

Mehrnaz-Charkhchi (approved):

veidenberg (approved):

jyo-nt (approved):

PR #27, opened 2024-08-01
Update Dockerfile to create temp dir.

Mehrnaz-Charkhchi:
### Description
Create a new directory in docker container to map to the actual /nfs directory accessible to nextflow pipelines. 

### Related JIRA Issue(s)
[ENSWBSITES-2691](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2691)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ COPY ./requirements.txt requirements.txt
 RUN pip install --no-cache-dir --upgrade -r requirements.txt
 COPY ./app /app
 COPY ./data /data
-RUN mkdir -p /nfs/public/rw/enswbsites/tmpdir
+RUN mkdir -p /nfs/public/rw/enswbsites
 RUN mkdir /tmpdir
 ENV TMPDIR /tmpdir
 ENV PORT 8013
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -4,6 +4,7 @@ COPY ./requirements.txt requirements.txt
 RUN pip install --no-cache-dir --upgrade -r requirements.txt
 COPY ./app /app
 COPY ./data /data
+RUN mkdir -p /nfs/public/rw/enswbsites/tmpdir
```
azangru: I have a couple of questions (sorry; that's because I don't know either Docker or our file systems):

- Do you actually need to create the directory manually? Won't Docker create the directory on its own when you mount the volume (e.g. [StackOverflow link](https://stackoverflow.com/a/24641998/3925302))
- Where can we see the contents of the `enswbsites` directory outside of the container? I can see `/nfs/public/rw/enswbsites/` only inside the datamover; and it contains four directories: `dev  internal  production  staging`. The reason I am asking is, do we need some dedicated directory inside of `enswbsites` specifically for vep files; or is a temporary directory at the top of `enswbsites` good enough?

Mehrnaz-Charkhchi: - Docker  does create a directory when we're mounting but it might not be the case in some distributions of Linux. That's why I'm creating it here to be on the safe side and to have a better understanding of what is happening. It doesn't cost anything to have this here and we can also use it for development env.
- If you look at the other PRs the actual directory that is used for different ens is /nfs/public/rw/enswbsites/staging/vep ...

PR #29, opened 2024-08-09
Update VEP config

veidenberg:
### Description
Update VEP config to support unsorted input and use the new workflow repo (+ cleanup of unused nextflow payload params). 

### Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2709

### DIFF
--- a/app/vep/models/pipeline_model.py
+++ b/app/vep/models/pipeline_model.py
@@ -21,7 +21,7 @@ class VEPConfigParams(BaseModel):
     vep_config: str
     bin_size: int = 3000
     sort: bool = True
-    output_prefix: str = ""  # optional prefix for output files
+    output_prefix: str = ""
 
     @model_serializer
     def vep_config_serialiser(self):
--- a/app/vep/models/pipeline_model.py
+++ b/app/vep/models/pipeline_model.py
@@ -49,7 +49,6 @@ class ConfigIniParams(BaseModel):
     transcript_version: int = 1
     canonical: int = 1
     sort: int = 1
-    # TODO: use genome_id from client payload to set gff+fasta paths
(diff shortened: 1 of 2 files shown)


app/vep/models/pipeline_model.py
```diff
@@ -55,11 +48,13 @@ class ConfigIniParams(BaseModel):
     biotype: bool = False
     transcript_version: int = 1
     canonical: int = 1
-    gff: str = "/nfs/public/rw/enswbsites/dev/vep/test/genes.gff3.bgz"
+    sort: int = 1
+    # TODO: use genome_id from client payload to set gff+fasta paths
```
Mehrnaz-Charkhchi: We have already discussed these hardcoded parts, Are these comments necessary? is there an actual update here that I'm missing?

veidenberg: Just a reminder that dynamic paths need to be implemented ~~for beta-2 (i.e. in this sprint?)~~. The comments are not the update but the changes in VEP config in order to work with the new pipeline (goes with the other PRs; see the PR description and linked JIRA ticket).

azangru: > dynamic paths need to be implemented for beta-2

Erm, isn't beta2 happening earlier/independently of VEP?

> (i.e. in this sprint?)

We would need loaded production data, an up-to-date grpc service probably pointing at the updated database, plus updates to the web-metadata-api that will pick the new attributes from the grpc response. Sounds unlikely for this sprint.

veidenberg: Yes, you're right, sorry for the confusion. No need to worry about it now. Comments removed.

PR #30, opened 2024-08-09
Update to vcf_results to support vcf.gz files

jonk-ebi:
### Description
Update to vcf_results.get_results_from_path to support vcf.gz files

### Related JIRA Issue(s)
[ENSWBSITES-2704](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2704)

### Review App URL(s) 
N/A

### Knowledge Base
This update also resolves a regression introduced during the restructuring of the repo. Unit tests for vcf will now run

### Example(s)
An easy way to test this change is to remove the skip from `test_get_results_with_file_and_dump` and update the `vcf_path` to a local file. This will load in the file and create a json dump of the expected page. Test expects a full page of 100 variants for page 2 so if your test file does not include >300 variants you will need to change this `assert`

### Checklist

- [x] Black formatting
- [x] Tests

### Dependencies 

If you get any errors about headers like `length` make sure you are using `git+https://github.com/likhitha-surapaneni/vcfpy@header-fix` for vcfpy

### DIFF
--- a/app/vep/utils/vcf_results.py
+++ b/app/vep/utils/vcf_results.py
@@ -52,9 +52,8 @@ def _set_allele_type(alt_one_bp: bool, ref_one_bp: bool, ref_alt_equal_bp: bool)
 
 
 def _get_prediction_index_map(
-                        csq_header: str,
-                        target_columns: List[str] = None
-                    ) -> Dict:
+    csq_header: str, target_columns: List[str] = None
+) -> Dict:
     """Creates a dictionary of column indexes based
     on the CSQ info description"""
     if not target_columns:
@@ -102,7 +101,7 @@ def _get_alt_allele_details(
         if len(cons) == 0:
             cons = []
         else:
-            cons = cons.split('&')
+            cons = cons.split("&")
         if csq_values[index_map["Feature_type"]] == "Transcript":
             is_cononical = (
                 _get_csq_value(csq_values, "CANONICAL", "NO", index_map) == "YES"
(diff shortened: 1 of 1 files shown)


app/vep/utils/vcf_results.py
```diff
@@ -229,10 +239,8 @@ def get_results_from_stream(
...
         metadata=model.Metadata(
             pagination=model.PaginationMetadata(
-                page=page,
-                per_page=page_size,
-                total=total
+                page=page, per_page=page_size, total=total
             )
         ),
-        variants=variants
+        variants=variants,
```
jyothishnt, line 245:
```suggestion
        variants=variants
```

app/vep/utils/vcf_results.py
```diff
@@ -1,10 +1,11 @@
 """ Module for loading a VCF and parsing it into a VepResultsResponse
 object as defined in APISpecification"""
 
+from typing import List, Dict, Any, IO
+
 import vcfpy
-from typing import List, Dict, Any
 
-from vep.models import vcf_results_model as model
+from app.vep.models import vcf_results_model as model
```
jyothishnt, line 8: do you need `app.`? complained for me locally.

jyothishnt: Aah! yes to run the test you need to access it using `app.`  but not for the app.

app/vep/utils/vcf_results.py
jonk-ebi: Black adds the trailing `,`. Python is happy with it and it is believed that it is nicer for anyone extending the code to add additional elements

app/vep/utils/vcf_results.py
jonk-ebi: Yeah I needed that and IO before the unit tests would run

PR #31, opened 2024-08-09
VEP results endpoint

jyothishnt:
### Description
This PR adds `/results` endpoint for the client to retrieve VEP results
This was tested using a sample vcf file after copying it to docker locally

Screenshot 2024-08-09 at 14 26 41

### Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2704

### Review App URL(s) 
http://vep-results-endpoint.review.ensembl.org

### Dependencies 
https://github.com/Ensembl/ensembl-web-tools-api/pull/30

### DIFF
--- a/app/vep/vep_resources.py
+++ b/app/vep/vep_resources.py
@@ -155,11 +155,12 @@ async def fetch_results(request: Request, submission_id: str, page: int, page_si
                     "details": f"A submission with id {submission_id} was not found",
                 }
             )
-            return PlainTextResponse(
-                response_msg, status_code=status.HTTP_404_NOT_FOUND
+            return JSONResponse(
+                content=response_msg, status_code=status.HTTP_404_NOT_FOUND
             )
-        return response_error_handler(result={"status": http_error.response.status_code})
+        return response_error_handler(
+            result={"status": http_error.response.status_code}
+        )
     except Exception as e:
         logging.debug(e)
-        # return response_error_handler(result={"status": 500})
-        return PlainTextResponse(e, status_code=500)
+        return response_error_handler(result={"status": 500})
--- a/app/vep/vep_resources.py
+++ b/app/vep/vep_resources.py
@@ -21,6 +21,7 @@ import logging
 from requests import HTTPError
 from starlette.responses import JSONResponse, PlainTextResponse, FileResponse
 from fastapi import Request, status, APIRouter
+from enum import Enum
 
 from core.error_response import response_error_handler
 from core.logging import InterceptHandler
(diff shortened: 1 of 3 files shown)


app/vep/vep_resources.py
```diff
@@ -140,3 +141,30 @@ async def download_results(request: Request, submission_id: str):
     except Exception as e:
         logging.debug(e)
         return response_error_handler(result={"status": 500})
+
+@router.get("/submissions/{submission_id}/results", name="view_results")
+async def fetch_results(request: Request, submission_id: str, page: int, page_size: int):
```
azangru: In the search api, we named the parameter that is similar to the `page_size` parameter here, `per_page` ([link](https://gitlab.ebi.ac.uk/ensembl-web/ensembl_search_hub/-/blob/master/APISpecification.yaml?ref_type=heads#L85-105)). Is it worth using the same convention in this endpoint as well?

azangru: Also, pages should start at 1, not at 0, as in your screenshot

image

jonk-ebi: This is down to the vcf_results module. @jyothishnt do you want me to make a change in this pr to resolve it?

app/vep/vep_resources.py
```diff
@@ -140,3 +141,30 @@ async def download_results(request: Request, submission_id: str):
...
+    try:
+        workflow_status = await get_workflow_status(submission_id)
+        submission_status = PipelineStatus(submission_id=submission_id, status=workflow_status['workflow']['status'])
+        if submission_status.status == "SUCCEEDED":
```
jonk-ebi: It would be good if we made a enum for status values rather than hardcoded strings. You get auto complete, if the string changes you only have to make one change, reduces the chance of typos, and it makes it explicit to future developers what statuses the service may have to deal with.

app/vep/vep_resources.py
jyo-nt: yeah @jonk-ebi if you could, please commit to this branch.

kamaldodiya (commented):
I think first the results-download should be merged and also there are quite few changed in results-download branch which are not here that needs to be updated as well. Also there is a chain of interdependent PRs

kamaldodiya (changes requested):
Piuckup commits from #32

kamaldodiya (approved):

azangru (approved):
Technically, there should probably be an else after the if on line 152? But web client shouldn't send requests for jobs that haven't succeeded.

PR #33, opened 2024-08-15
Catch unfinished job on results endpoint

jyothishnt:
This is just a missed piece of code from vep results endpoint.


kamaldodiya (approved):
Looks good but shouldn't be merged until the VEP import issue is fixed

PR #35, opened 2024-08-16
Update VEP config

veidenberg:
### Description
Update VEP config + pipeline params to switch Tools API to new VEP workflow from Variation.

### Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2709

### Dependencies
Requires config updated for this to work which is PR https://gitlab.ebi.ac.uk/ensembl-web/newsite-api-config-setup/-/merge_requests/15

### DIFF
--- a/app/vep/models/pipeline_model.py
+++ b/app/vep/models/pipeline_model.py
@@ -13,7 +13,7 @@ from core.config import (
     NF_WORK_DIR,
 )
 from core.logging import InterceptHandler
-import logging, os
+import json, logging, os
 
 logging.getLogger().handlers = [InterceptHandler()]
 
@@ -24,7 +24,6 @@ class VEPConfigParams(BaseModel):
     outdir: DirectoryPath
     bin_size: int = 3000
     sort: bool = True
-    output_prefix: str = ""
 
     @model_serializer
     def vep_config_serialiser(self):
@@ -33,12 +32,9 @@ class VEPConfigParams(BaseModel):
         outdir_str = f'"outdir": "{self.outdir.as_posix()}"'
         bin_str = f'"bin_size": {self.bin_size}'
         sort_str = f'"sort": {"true" if self.sort else "false"}'
-        prefix_str = (
-            f'"output_prefix": "{self.output_prefix}"' if self.output_prefix else ""
-        )
         json_str = (
             "{" + ", ".join(
-                [vcf_str, config_str, outdir_str, bin_str, sort_str, prefix_str]
+                [vcf_str, config_str, outdir_str, bin_str, sort_str]
             ) + "}"
         )
         return json_str
--- a/app/vep/models/pipeline_model.py
+++ b/app/vep/models/pipeline_model.py
@@ -33,11 +33,14 @@ class VEPConfigParams(BaseModel):
         outdir_str = f'"outdir": "{self.outdir.as_posix()}"'
         bin_str = f'"bin_size": {self.bin_size}'
         sort_str = f'"sort": {"true" if self.sort else "false"}'
-        prefix_str = f'"output_prefix": "{self.output_prefix}"' if self.output_prefix else ""
-        json_str = "{" + ", ".join(
-            [vcf_str, config_str, outdir_str, bin_str, sort_str, prefix_str]
-        ) + "}"
-
+        prefix_str = (
+            f'"output_prefix": "{self.output_prefix}"' if self.output_prefix else ""
+        )
+        json_str = (
+            "{" + ", ".join(
+                [vcf_str, config_str, outdir_str, bin_str, sort_str, prefix_str]
+            ) + "}"
+        )
         return json_str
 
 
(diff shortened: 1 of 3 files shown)


app/vep/models/pipeline_model.py
```diff
@@ -1,4 +1,3 @@
-from typing import List
```
veidenberg, line 1: Not needed since python 3.9 (we're using 3.10)

app/vep/models/pipeline_model.py
```diff
@@ -91,7 +93,7 @@ def create_config_ini_file(self, dir):
 class PipelineStatus(BaseModel):
     submission_id: str
     status: str = Field(
-        alias=AliasPath("status", "workflow", "status"), default="FAILED"
```
veidenberg, line 94: `AliasPath` can only be used with `validation_alias` (`alias` expects a string)

azangru (approved):

app/vep/models/pipeline_model.py
Mehrnaz-Charkhchi: I don't think this is the case. We use aliaspath the same way on metadata and it's working.

app/vep/models/pipeline_model.py
```diff
@@ -23,16 +22,26 @@ class VEPConfigParams(BaseModel):
...
+        prefix_str = (
+            f'"output_prefix": "{self.output_prefix}"' if self.output_prefix else ""
         )
-        return stringified_encoded_json
+        json_str = (
+            "{" + ", ".join(
+                [vcf_str, config_str, outdir_str, bin_str, sort_str, prefix_str]
+            ) + "}"
+        )
+        return json_str
```
kamaldodiya, line 40: Did you test the file being generated here and see if it actually works with the VEP ?
Alternatively, did you compare it with the current working file.

app/vep/models/pipeline_model.py
kamaldodiya: Agree with @Mehrnaz-Charkhchi  alias, aliaspath is just the mechanism to extract the value.

app/vep/models/pipeline_model.py
kamaldodiya: I am getting 400 error while launching the workflow. 
```
  File "/Users/kamal/Development/venvs/tad310/lib/python3.10/site-packages/requests/models.py", line 1024, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
          │         │                        └ <Response [400]>
          │         └ '400 Client Error: Bad Request for url: https://api.cloud.seqera.io/workflow/launch?workspaceId=185194778628596'
          └ <class 'requests.exceptions.HTTPError'>

```

kamaldodiya (commented):
Do we want to change the default pipeline in the env.sample as well 
https://github.com/Ensembl/ensembl-web-tools-api/blob/dev/app/.env.sample#L4

Mehrnaz-Charkhchi:
> Do we want to change the default pipeline in the env.sample as well https://github.com/Ensembl/ensembl-web-tools-api/blob/dev/app/.env.sample#L4

It would be good if we could try testing the codes we update before pushing it to live because even if we write tests all of these parts are going to be mocked so we can never know!

app/vep/models/pipeline_model.py
veidenberg: Unless I'm missing something, `AliasPath` can be used to extract information, but not to serialize it (according to documentation and MyPy reports), so the correct way is to use `validation_alias` instead of `alias` param. Same with metadata.

app/vep/models/pipeline_model.py
veidenberg: I could not reproduce it when running ToolsAPI on my local machine (got job ID back when launching the workflow with payload from PR #22). Where does this 400 come from?

veidenberg: Edit: I could reproduce the error when changing the serializer output, e.g. adding another json.dumps() around it. So the current code should work fine.

kamaldodiya: This works required config update

app/vep/models/pipeline_model.py
```diff
@@ -23,16 +22,26 @@ class VEPConfigParams(BaseModel):
...
+        outdir_str = f'"outdir": "{self.outdir.as_posix()}"'
+        bin_str = f'"bin_size": {self.bin_size}'
+        sort_str = f'"sort": {"true" if self.sort else "false"}'
+        prefix_str = (
+            f'"output_prefix": "{self.output_prefix}"' if self.output_prefix else ""
         )
-        return stringified_encoded_json
+        json_str = (
+            "{" + ", ".join(
+                [vcf_str, config_str, outdir_str, bin_str, sort_str, prefix_str]
```
kamaldodiya: prefix_str being empty it might invalidate json.

kamaldodiya (changes requested):

kamaldodiya (approved):

PR #36, opened 2024-08-16
VEP: update logic for generating alt alleles in the vcf to json script

azangru:
### Bug
In response from https://staging-2020.ensembl.org/api/tools/vep/submissions/5UvdiTnQBeK0gU/results?page=1&per_page=100, the allele_type of alternative alleles is spelled incorrectly ("INS" rather than "insertion"); and the `predicted_molecular_consequences` array is empty, which should never happen:

```
{
  "metadata": {
    "pagination": {
      "page": 1,
      "per_page": 100,
      "total": 1
    }
  },
  "variants": [
    {
      "name": "rs1405511870",
      "allele_type": "insertion",
      "location": {
        "region_name": "1",
        "start": 924510,
        "end": 924511
      },
      "reference_allele": {
        "allele_sequence": "C"
      },
      "alternative_alleles": [
        {
          "allele_sequence": "CGCTGCC",
          "allele_type": "INS",
          "representative_population_allele_frequency": null,
          "predicted_molecular_consequences": []
        }
      ]
    }
  ]
}
```

See file from which this json was generated ([Slack link](https://genomes-ebi.slack.com/archives/G0104HU9R1C/p1723846118817199))

### Cause
The code was comparing sequences from the `ALT` field of the vcf file with allele sequences of the `CSQ` data (see vcf file). However, for some allele types, such as insertions, indels, and probably deletions, these strings may differ because of different conventions regarding the inclusion of the "anchor" nucleotide into the allele sequence.

### DIFF
--- a/app/vep/utils/vcf_results.py
+++ b/app/vep/utils/vcf_results.py
@@ -81,17 +81,20 @@ def _get_csq_value(
 
 
 def _get_alt_allele_details(
-    alt: vcfpy.AltRecord, csqs: List[str], index_map: Dict
+    alt: str, ref: str, csqs: List[str], index_map: Dict
 ) -> model.AlternativeVariantAllele:
     """Creates  AlternativeVariantAllele based on
     target alt allele and CSQ entires"""
     frequency = None
     consequences = []
+    allele_type = _set_allele_type(
+        len(alt) < 2, len(ref) < 2, len(alt) == len(ref)
+    )[0]
 
     for str_csq in csqs:
         csq_values = str_csq.split("|")
 
-        if csq_values[index_map["Allele"]] != alt.value:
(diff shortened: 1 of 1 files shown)


app/vep/utils/vcf_results.py
```diff
@@ -152,7 +155,7 @@ def get_results_from_path(
 
 
 def get_results_from_stream(
-    page_size: int, page: int, vcf_stream: IO
```
azangru, line 155: VSCode complained to me about IO. I don't know if it is a valid type.

veidenberg (approved):

jonk-ebi (approved):
Change makes sense. With a little work all unit tests passed.

PR #37, opened 2024-09-09
Dynamically generate VEP support paths.

Mehrnaz-Charkhchi:
### Description
Update the tools API so the gff and fasta entries in ConfigIniParams are dynamically generated based on metadata values

### Related JIRA Issue(s)
[ENSWBSITES-2726](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2726)

### Review App URL(s) 

### Checklist

- [*] Black formatting
- [ ] Tests

### DIFF
--- a/app/.env.sample
+++ b/app/.env.sample
@@ -5,3 +5,5 @@ NF_PIPELINE_URL=https://github.com/jonk-ebi/platform-test-pipeline
 NF_WORK_DIR=/path/to/work/dir
 SEQERA_API=https://api.cloud.seqera.io
 UPLOAD_DIRECTORY=/tmpdir
+WEB_METADATA_API=https://beta.ensembl.org/api/metadata/
+VEP_SUPPORT_PATH=/nfs/public/ro/enswbsites/dev/vep/support/
--- a/app/core/config.py
+++ b/app/core/config.py
@@ -53,12 +53,12 @@ for logger_name in LOGGERS:
 logger.configure(handlers=[{"sink": sys.stderr, "level": LOGGING_LEVEL}])
 
 # Nextflow Configurations
-NF_TOKEN = config("NF_TOKEN", default="")
-NF_COMPUTE_ENV_ID = config("NF_COMPUTE_ENV_ID", default="")
-NF_PIPELINE_URL = config("NF_PIPELINE_URL", default="")
-NF_WORK_DIR = config("NF_WORK_DIR", default="")
-SEQERA_API = config("SEQERA_API", default="")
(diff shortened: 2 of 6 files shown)


app/.env.sample
```diff
@@ -5,3 +5,5 @@ NF_PIPELINE_URL=https://github.com/jonk-ebi/platform-test-pipeline
 NF_WORK_DIR=/path/to/work/dir
 SEQERA_API=https://api.cloud.seqera.io
 UPLOAD_DIRECTORY=/tmpdir
+WEB_METADATA_API=https://beta.ensembl.org/api/metadata/
+VEP_SUPPORT_PATH=/nfs/public/ro/enswbsites/dev/vep/support/
```
azangru, line 9: Why do we call it a "support" path? This name doesn't really tell much. VEP data files path? Or, if this is only for genomic data, then perhaps VEP genebuild path or VEP assembly path?

It might be worth asking Variation what other file types they expect we will need in the future; and asking production whether those files would be stored together with these gff and fasta files, or in some different directory.

kamaldodiya: Based on the code review discussion. This is good enough for VEP phase 1. Will need to revisit this once we have different files and how production is going to save it and how web is going to access it.

app/core/config.py
```diff
@@ -39,27 +39,29 @@
...
-NF_PIPELINE_URL = config("NF_PIPELINE_URL")
-NF_WORK_DIR = config("NF_WORK_DIR")
-SEQERA_API = config("SEQERA_API")
-NF_WORKSPACE_ID = config("NF_WORKSPACE_ID")
+NF_TOKEN = config("NF_TOKEN", default="")
+NF_COMPUTE_ENV_ID = config("NF_COMPUTE_ENV_ID", default="")
+NF_PIPELINE_URL = config("NF_PIPELINE_URL", default="")
+NF_WORK_DIR = config("NF_WORK_DIR", default="")
+SEQERA_API = config("SEQERA_API", default="")
+NF_WORKSPACE_ID = config("NF_WORKSPACE_ID", default="")
```
kamaldodiya: Without all these variables VEP won't work. Is setting default to "" a good idea ?

Mehrnaz-Charkhchi: You're right. I'll move them to docker-compose.

kamaldodiya: That will serve the purpose. Also as it is already set in the .env file. Exposing .env file to docker-compose would work for both docker based env/ virtualenv/production across alll the environment dev/prod etc.

kamaldodiya (approved):

azangru (approved):

jyobhai (approved):

PR #38, opened 2024-09-09
new vep config endpoint

jyobhai:
### Description
New vep form config endpoint

### Related JIRA Issue(s)
https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2647

### Review App URL(s) 
http://vep-form-config.review.ensembl.org

### Checklist

- [x] Black formatting
- [  ] Tests


app/vep/utils/web_metadata.py
```diff
@@ -0,0 +1,26 @@
...
+          metadata = response.json()
+    except requests.exceptions.HTTPError as HTTPError:
+      logger.error(f"HTTPError: {HTTPError}")
+      return None
+    except Exception as e:
+      logger.exception(e)
+      return None
+
+    provider = metadata["annotation_provider"]
+    version = metadata["annotation_version"]
```
azangru: We may not be ready to use the annotation_version yet (Daniel promised to update it with proper data; but right now it may contain the "internal version" not intended for the public). The fallback, I believe, is to use the last updated date.

app/vep/utils/web_metadata.py
```diff
@@ -0,0 +1,26 @@
...
+from typing import List
+from core.config import ENSEMBL_WEB_METADATA_API
+from vep.models.client_model import GenomeAnnotationProvider
+
+async def get_genome_metadata(genome_id: str) -> GenomeAnnotationProvider:
+
+    try:
+      session = requests.Session()
+      with session.get(
+          url=f"https://staging-2020.ensembl.org/api/metadata/genome/{genome_id}/details"
```
azangru: Part of this url string should come from an environment variable

jyothishnt: Yes it was! but changed it for testing and forgot to undo. 👍

app/core/config.py
```diff
@@ -63,3 +63,4 @@
 NF_WORKSPACE_ID = config("NF_WORKSPACE_ID")
 
 UPLOAD_DIRECTORY = config("UPLOAD_DIRECTORY", default="/tmpdir")
+ENSEMBL_WEB_METADATA_API = config("ENSEMBL_WEB_METADATA_API")
```
Mehrnaz-Charkhchi: Checkout my PR for this env variable WEB_METADATA_API.

app/vep/models/client_model.py
```diff
@@ -1 +1,39 @@
+from pydantic import BaseModel, Field
+from typing import List, Optional
 
+def default_options():
```
Mehrnaz-Charkhchi: Do we need this ?

app/vep/utils/web_metadata.py
```diff
@@ -1,26 +1,32 @@
...
+        response.raise_for_status()
+        attributes = {}
+        print (attributes)
+        for attribute in response.json()['attributes']:
+            name = attribute['name']
+            value = attribute['value']
+            attributes[name] = value
+        return attributes
+    except HTTPError as http_error:
+        logging.debug(http_error)
```
azangru: Could you ask @Mehrnaz-Charkhchi what logging level will be reasonable for this and other places where you use it? I think here, for instance, it should at least be a logging.warn.

app/vep/utils/web_metadata.py
```diff
@@ -1,26 +1,32 @@
...
+            + "genome/"
+            + genome_id
+            + "/dataset/genebuild/attributes?"
+            + "attribute_names=genebuild.provider_name&"
+            + "attribute_names=genebuild.display_version&"
+            + "attribute_names=genebuild.last_geneset_update"
+        )
+        response.raise_for_status()
+        attributes = {}
+        print (attributes)
```
azangru: we probably don't want this?

app/vep/models/client_model.py
jyothishnt: yes, to add default options on line 17

app/core/config.py
jyothishnt: Would you like me to use the same name?

Mehrnaz-Charkhchi: Yes please, It makes sense to use the same config name.

app/vep/models/client_model.py
Mehrnaz-Charkhchi: Right, We usually add constant variables.
https://stackoverflow.com/questions/77561280/const-pydantic-field-value

app/vep/utils/web_metadata.py
Mehrnaz-Charkhchi: @jyobhai We should use logging.error for errors that we want to raise.

app/vep/vep_resources.py
```diff
@@ -209,3 +211,38 @@ async def fetch_results(request: Request, submission_id: str, page: int, per_pag
...
+                }
+            )
+            return JSONResponse(
+                content=response_msg, status_code=status.HTTP_404_NOT_FOUND
+            )
+        return response_error_handler(
+            result={"status": http_error.response.status_code}
+        )
+    except Exception as e:
+        logging.debug(e)
```
Mehrnaz-Charkhchi: Same here. logging.error.

app/vep/models/client_model.py
jyobhai: Can we discuss this?

app/vep/models/client_model.py
```diff
@@ -1 +1,38 @@
+from pydantic import BaseModel, Field
+from typing import List, Optional
 
+class FieldOptions:
+    GENCODE_LABEL: str = "GENCODE"
+    GENCODE_VALUE: str = "gencode_comprehensive"
```
azangru: Sorry, I am very confused about the naming of these. Why are we calling them `GENCODE_LABEL` or `GENCODE_VALUE`. "Gencode" is the annotation provider just for human. This should probably called `TRANSCRIPT_SET_LABEL` and `TRANSCRIPT_SET_VALUE`; and it doesn't make sense to have default values for them, because they are genome-specific.

jyothishnt: Hi @azangru , you are right, I have to remove the GENCODE part from that key. 
There is a label inside transcript_set and a label inside options. I think I did it wrong, I updated the `transcript_set.label` with dynamic data from grpc instead of `options`. This is wrong I guess?

app/vep/models/client_model.py
```diff
@@ -1 +1,38 @@
...
+class Options(BaseModel):
+  label: str = FieldOptions.GENCODE_LABEL
+  value: str = FieldOptions.GENCODE_VALUE
+
+class TranscriptSet(BaseModel):
+    label: str
+    description: str = None
+    type: str = "select"
+    options: List[Options] = [Options()]
+    default_value: str = FieldOptions.GENCODE_VALUE
```
azangru: Transcript set is genome-specific; so it doesn't make sense to have a default value

app/vep/models/client_model.py
```diff
@@ -1 +1,38 @@
+from pydantic import BaseModel, Field
+from typing import List, Optional
 
+class FieldOptions:
+    GENCODE_LABEL: str = "GENCODE"
+    GENCODE_VALUE: str = "gencode_comprehensive"
+
+class Options(BaseModel):
+  label: str = FieldOptions.GENCODE_LABEL
+  value: str = FieldOptions.GENCODE_VALUE
```
azangru: Initialise to an empty string rather than to `FieldOptions.GENCODE_LABEL` / `FieldOptions.GENCODE_VALUE`?

app/vep/models/client_model.py
azangru: It seems that this whole class is unnecessary.

app/vep/models/client_model.py
```diff
@@ -1 +1,38 @@
...
+
+class TranscriptSet(BaseModel):
+    label: str
+    description: str = None
```
azangru: `str` | `None` ?

app/vep/models/client_model.py
```diff
@@ -1 +1,38 @@
...
+    type: str = "select"
+    options: List[Options] = [Options()]
+    default_value: str = FieldOptions.GENCODE_VALUE
+
+class Symbol(BaseModel):
+    label: str = "Gene symbol"
+    description: str = None
```
azangru: str | None

azangru: There should be a class like `BooleanOption` (or `CheckboxOption`) which will describe the type. Both symbol and biotype are instances of that class; they do not need their own classes.

app/vep/models/client_model.py
```diff
@@ -1 +1,38 @@
...
 
+class FieldOptions:
+    GENCODE_LABEL: str = "GENCODE"
+    GENCODE_VALUE: str = "gencode_comprehensive"
+
+class Options(BaseModel):
+  label: str = FieldOptions.GENCODE_LABEL
+  value: str = FieldOptions.GENCODE_VALUE
+
+class TranscriptSet(BaseModel):
```
azangru: This should probably be something like a `SelectOption` class. There will be other options of this type in the future, I am sure.

app/vep/vep_resources.py
```diff
@@ -210,3 +212,38 @@ async def fetch_results(request: Request, submission_id: str, page: int, per_pag
...
+            label=f"{annotation_provider_name} {annotation_version or last_updated_date}",
+        )
+
+        form_config = FormConfig(
+            transcript_set = transcript_set
+        )
+        return form_config
+
+    except HTTPError as http_error:
+        if http_error.response.status_code in [403, 400]:
```
azangru: Limiting to `403, 400` is probably incorrect?

What should happen after the exception is caught? Probably just log the error and send the same response to client?

app/vep/models/client_model.py
jyothishnt: Isn't `default_value` that tells you the default from a list of options?

azangru: Oh, sorry, my bad, you are right. The transcript set, for the foreseeable future, will only contain a single option, and the default value should be the value of that option.

Since the client won't send transcript set's value back to the server (at least not unless we introduce the ability to choose between different transcript sets), it does not _really_ matter what the actual value is. It could be an empty string. Or it could be a slightly more meaningful string such as "default_transcript_set" or just "default". Or it could be generated dynamically from the provider name, version, and date, which is probably an overkill for now.

jyothishnt: I am generating the label with provider name and version. So I can use the same for `value` and `default_value`

app/vep/utils/web_metadata.py
```diff
@@ -24,3 +25,27 @@ def get_vep_support_location(genome_id):
...
+        for attribute in response.json()['attributes']:
+            name = attribute['name']
+            value = attribute['value']
+            attributes[name] = value
+        return attributes
+    except HTTPError as http_error:
+        logging.error(http_error)
+        raise HTTPError
+    except Exception as e:
+        logging.error(e)
```
veidenberg, line 50: Follow-up to code reaview: please add `raise Excpetion` here (I can't add it myself as I don't have this function in [my PR](https://github.com/Ensembl/ensembl-web-tools-api/pull/42)). Re-raising allows the caller (`get_form_config()`) to handle the original Exception instead of handling the unexpected `None` return value [here](https://github.com/Ensembl/ensembl-web-tools-api/blob/636ac1c9e28f2476074bab536f034b5847db9efb/app/vep/vep_resources.py#L220).

veidenberg (approved):

app/vep/utils/web_metadata.py
```diff
@@ -1,13 +1,14 @@
+from loguru import logger
 import requests
-
-import logging
 from requests import HTTPError
-
-from core.config import WEB_METADATA_API, VEP_SUPPORT_PATH
+from typing import List
```
veidenberg: not needed (you can use built-in `list` type instead), but also not blocking PR merge

azangru (approved):

kamaldodiya (approved):

Mehrnaz-Charkhchi (approved):

PR #41, opened 2024-09-12
Python logging

Mehrnaz-Charkhchi:
### Description
Fix python logging.

### Review App URL(s) 

### Checklist

- [ ] Black formatting
- [ ] Tests

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.10
+FROM python:3.11
 WORKDIR /app
 COPY ./requirements.txt requirements.txt
 RUN pip install --no-cache-dir --upgrade -r requirements.txt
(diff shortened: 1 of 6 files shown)


app/core/config.py
```diff
@@ -41,10 +41,10 @@
 
 
 # logging configuration
-
+logging.basicConfig(level=logging.DEBUG)
 LOGGING_LEVEL = logging.DEBUG if DEBUG else logging.INFO
 LOGGERS = ("uvicorn.asgi", "uvicorn.access")
-
+log = logging.getLogger("gunicorn.access")
```
kamaldodiya: Is it `uvicorn.access` or `gunicorn.access` ?

Dockerfile
```diff
@@ -7,5 +7,6 @@ COPY ./data /data
 RUN mkdir -p /nfs/public/rw/enswbsites
 ENV TMPDIR /nfs/public/rw/enswbsites
 ENV PORT 8013
-EXPOSE 8013
+EXPOSE 8018
```
kamaldodiya: 8013 ?

kamaldodiya (changes requested):

app/core/config.py
Mehrnaz-Charkhchi: I removed it but it's in most of our projects just in case we start using gunicorn but it doesn't have an impact otherwise.

Dockerfile
Mehrnaz-Charkhchi: Updated.

PR #42, opened 2024-09-20
Update agent workdir

veidenberg:
### Description
This PR sets the workdir param for each VEP job submission to the (tempdir) input file directory to keep all the job files together.

Example jobdir with the new workdir param ([link to submission](https://cloud.seqera.io/orgs/EMBL-EBI/workspaces/ebi-spws-dev-ensembl-next-tools/watch/Bbdo92UNqNz8c)):
```bash
[ens_adm02@hh-wp-ensembl-01 ~]$ ls /nfs/public/rw/enswbsites/dev/vep/tmp_pr_test
22  config.ini  e0  f7  input.txt  input_VEP.vcf.gz  input_VEP.vcf.gz.csi  singularity  vep-summary
# old job dir content for rerefernce
[ens_adm02@hh-wp-ensembl-01 ~]$ ls /nfs/public/rw/enswbsites/dev/vep/tmpmfsofykc
config.ini  input.txt  input_VEP.vcf.gz  input_VEP.vcf.gz.csi  vep-summary
```

### Related JIRA Issue(s)
[ENSWBSITES-2736](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2736)

### Review App URL(s) 
http://update-agent-wdir.review.ensembl.org (does not work, addressed in [ENSWBSITES-2759](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2759)).

### Checklist

- [x] Black formatting
- [x] Tests

### Dependencies 
Updated envvars in k8s ([merge request here](https://gitlab.ebi.ac.uk/ensembl-web/newsite-api-config-setup/-/merge_requests/21)).

###


app/vep/utils/web_metadata.py
```diff
@@ -24,3 +24,4 @@ def get_vep_support_location(genome_id):
         raise HTTPError
     except Exception as e:
         logging.error(e)
+        raise Exception
```
Mehrnaz-Charkhchi: You're already in an exception, why are you raising another one? what are you expecting here?

app/vep/models/upload_vcf_files.py
```diff
@@ -25,16 +25,16 @@
...
-from core.config import UPLOAD_DIRECTORY
+from core.config import NF_WORK_DIR
 
 MAX_FILE_SIZE = 1024 * 1024 * 1024 * 2  # 2GB
 MAX_REQUEST_BODY_SIZE = MAX_FILE_SIZE + 1024
 
 
 class MaxBodySizeException(Exception):
-    def __init__(self, body_len: str):
+    def __init__(self, body_len: int):
```
Mehrnaz-Charkhchi, line 37: Please use str here.

Mehrnaz-Charkhchi (approved):

app/vep/utils/web_metadata.py
veidenberg: This was flagged by Pylance type checking: without re-raising, it will make this funciton to return `None`, which is not expected downstream, resulting in another, different Exception ([at this line](https://github.com/Ensembl/ensembl-web-tools-api/blob/994dc962d07672df2b5a3578e4b873159fe09f79/app/vep/models/pipeline_model.py#L69)).

app/vep/models/upload_vcf_files.py
veidenberg: Another issue flagged by Pylance: `MaxBodySizeException` is receiving `int` (e.g. [here](https://github.com/Ensembl/ensembl-web-tools-api/blob/994dc962d07672df2b5a3578e4b873159fe09f79/app/vep/models/upload_vcf_files.py#L49) and [here](https://github.com/Ensembl/ensembl-web-tools-api/blob/994dc962d07672df2b5a3578e4b873159fe09f79/app/vep/models/upload_vcf_files.py#L94)), but it was annotated as expecting `str`. Is there a reason?

kamaldodiya:
1. Review APP link is not working ?
2. Do you have a structure of the job directory after the changes made here ?

app/vep/models/upload_vcf_files.py
Mehrnaz-Charkhchi: The best practices were using str and that's why I would rather keep it that way.

veidenberg: Let's take a step back. The purpose of type hints is to catch typing errors, and `body_len: str` results in such an error. In this case, the type hint (expecting `str` values) conflicts with code (sending `int`s instead), so we only have two options to resolve it: change the type hint (like I've done here) or change the code around it (convert `body_len` param from `int` to `str` in all calls to `MaxBodySizeException`; see links in my comment above). I prefer correcting the type hint unless there's a good reason to go for (more complex) code modification route. I could not find a practice that recommends `str` type for `body_len` (aside few code sinppets like [this](https://github.com/fastapi/fastapi/discussions/6054) and [this](https://stackoverflow.com/questions/73442335/how-to-upload-a-large-file-≥3gb-to-fastapi-backend/73443824#73443824) which make the same mistake of using the wrong type hint). 
In short: I cannot just change the type hint back to `str`, because that would mean ignoring the typing error and the type hint itself (so why are we adding type hints at all if we choose to ignore it).

veidenberg:
> 1. Review APP link is not working ?

[Delpoyment](https://gitlab.ebi.ac.uk/ensembl-web/ensembl-tools-api/-/jobs/1811712) fails with `error validating "./": error validating data: ValidationError(Deployment.spec.template.spec.volumes[0].nfs): missing required field "path" in io.k8s.api.core.v1.NFSVolumeSource`. Not sure how to fix it: both the [template](https://gitlab.ebi.ac.uk/ensembl-web/ensembl-k8s-manifests/-/blob/k8s123-migration/ensembl-web-tools-api/deployment.yaml?ref_type=heads#L37) (the path field) and its [value](https://github.com/Ensembl/ensembl-web-tools-api/blob/0cbe56e54198008a6b6e0c854f36372f324f6b82/.gitlab-ci.yml#L21) ([HL_NFS_OPPATH_CODON
 envvar](https://gitlab.ebi.ac.uk/ensembl-web/ensembl-tools-api/-/settings/ci_cd)) are present.

> 2. Do you have a structure of the job directory after the changes made here ?

New structure: `/nfs/public/rw/enswbsites/[dev|production]/vep/<job_specific_tmp_dir>/<input/output/workflow files>` (previously input files were in job_tmp_dir but all intermediate/workflow files were scattered to `.../dev/vep`)
We can create an example once I get the review app deployed.

.gitlab-ci.yml
```diff
@@ -101,12 +101,14 @@ Review:
   before_script:
     - kubectl config use-context ${AGENT}
     # Check for review namespace and remove the default backend ingress
-    - kubectl get namespace ${NS} || (echo "Please create '${NS}' branch in ensembl-client and restart this pipeline" && exit 1)
+    - kubectl get namespace ${NS} > /dev/null || (echo "Please create '${NS}' branch in ensembl-client and restart this pipeline" && exit 1)
```
azangru: I am not sure I fully understand what is being proposed. Will every feature branch of the tools api need a corresponding client branch in order to get deployed properly? Are we not expecting review deployments that are just a bare back-end api?

veidenberg: Yes, (at least for Tools API) the assumption is that if the namespace is not there it means we have forgotten to create the client-side review app, and that's almost always the case:
* Tools API has no endpoint that is meant to be used independently from the client
* If Tools API would create the review namespace here, client deployment would fail (because it will try to create the same ns; hence the check here)
* We could add another job to deploy Tools API-only review app, but for what purpose? And how do we control which kind of review gets deployed?
* If we really want to create Tools API-only review app, we can still do it by creating the namespace manually before pushing the review branch.

Same approach is used in Genome Browser (other backends don't have review deployment option).

azangru: I get what you are saying; it just seems that there is something wrong with our review deployment setup that we need to switch to a different repo and create a new branch in order to create a review deployment for code in this repo. By doing so, we create extra clutter, which we then forget to clean up:

image

I feel like there should be some kind of a button, ideally with a bunch of associated selectors, that would configure and deploy the desired combination of branches of different services. I am surprised that there isn't a clean solution for this.

_(Not a blocker for the PR, of course)_

kamaldodiya:
As discussed in  the Code Review Meeting  @veidenberg   to log the payload and see if it has the appropriate paramsText, workDir Params and this PR should be good. 
@kamaldodiya  Tested this by sending the payload directly to seqera API and it is putting all the files to dedicated directory than the global one. 

Payload
```
{
    "launch": {
        "computeEnvId": "<COMPUTE_ENV>",
        "configProfiles": [
            "ensembl"
        ],
        "paramsText": "{\"input\": \"<BASE_PATH>/testdedicatedir/input.txt\", \"vep_config\": \"<BASE_PATH>/testdedicatedir/config.ini\", \"outdir\": \"<BASE_PATH>/testdedicatedir\", \"bin_size\": 3000, \"sort\": true}",
        "pipeline": "https://github.com/Ensembl/ensembl-vep",
        "pullLatest": true,
        "revision": "main",
        "workDir": "<BASE_PATH>/testdedicatedir/"
    }
}
```

Manual Test 
```
# Create Workflow 

curl -v -H "Authorization: Bearer <TOKEN>" -H "Content-Type: application/json" 'https://api.cloud.seqera.io/workflow/launch?workspaceId=<WS_ID>' --data @payload-dedicateddir-test.json

{"workflowId":"1R5vVxrXunFfc"}

# After Submission
$ testdedicatedir$ ls -l 
total 48
-rw-r--r-- 1 1234 5678 283 Sep 25 10:13 config.ini
-rw-r--r-- 1 1234 5678  17 Sep 25 10:13 input.txt

# After successful Run 
$ testdedicatedir$ ls -l 
total 224
drwxr-xr-x 4 1234 5678  96 Sep 25 10:23 59
drwxr-xr-x 3 1234 5678  48 Sep 25 10:23 c1
-rw-r--r-- 1 1234 5678 283 Sep 25 10:13 config.ini
-rw-r--r-- 1 1234 5678  17 Sep 25 10:13 input.txt
-rw-r--r-- 1 1234 5678 901 Sep 25 10:23 input_VEP.vcf.gz
-rw-r--r-- 1 1234 5678 112 Sep 25 10:23 input_VEP.vcf.gz.csi
drwxr-xr-x 2 1234 5678 120 Sep 25 10:23 singularity
drwxr-xr-x 2 1234 5678   0 Sep 25 10:23 vep-summary
```

veidenberg:
@kamaldodiya Payload JSON from local deployment:
```json
{
  "launch": {
    "computeEnvId": "<COMPUTE_ENV_ID>",
    "pipeline": "https://github.com/Ensembl/ensembl-vep",
    "workDir": "<WORK_DIR>/tmpdwv1fozj",
    "revision": "main",
    "pullLatest": true,
    "configProfiles": [
      "ensembl"
    ],
    "paramsText": "{\"input\": \"<WORK_DIR>/tmpdwv1fozj/small.vcf\", \"vep_config\": \"<WORK_DIR>/tmpdwv1fozj/config.ini\", \"outdir\": \"<WORK_DIR>/tmpdwv1fozj\", \"bin_size\": 3000, \"sort\": true}"
  }
}
```

kamaldodiya:
Ran this locally and there is a problem with the payload. The work directory needs to pre string in the payload.
```
{
  "launch": {
    "computeEnvId": "<COMPUTE_ENV_ID>",
    "pipeline": "https://github.com/Ensembl/ensembl-vep",
    "workDir": PosixPath("<WORK_DIR>/tmpdwv1fozj"),
    "revision": "main",
    "pullLatest": true,
    "configProfiles": [
      "ensembl"
    ],
    "paramsText": "{\"input\": \"<WORK_DIR>/tmpdwv1fozj/small.vcf\", \"vep_config\": \"<WORK_DIR>/tmpdwv1fozj/config.ini\", \"outdir\": \"<WORK_DIR>/tmpdwv1fozj\", \"bin_size\": 3000, \"sort\": true}"
  }
}
```

See the `workDir`  parameter which needs to be string. Adding serialiser would fix this issue.

kamaldodiya (approved):
Looks Good. It has conflicts which needs to be resolved and see the tests has passed or else it will fail in pipeline

PR #44, opened 2024-09-25
fix annotation provider details

jyothishnt:
If `annotation provider version` or `last updated date` is unavailable then update payload accordingly so that it doesnt return "None" to the client

### DIFF
--- a/app/vep/vep_resources.py
+++ b/app/vep/vep_resources.py
@@ -218,12 +218,19 @@ async def get_form_config(request: Request, genome_id: str):
     try:
         attributes = await get_genome_metadata(genome_id)
         annotation_provider_name = attributes.get("genebuild.provider_name")
-        annotation_version = attributes.get("genebuild.display_version")
-        last_updated_date = attributes.get("genebuild.last_geneset_update")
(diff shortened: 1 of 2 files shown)


azangru:
Is this a good idea? The details endpoint is very slow right now.

jyothishnt:
> Is this a good idea? The details endpoint is very slow right now.

I dont know, lets discuss in the daily scrum

app/vep/vep_resources.py
```diff
@@ -218,12 +218,19 @@ async def get_form_config(request: Request, genome_id: str):
     try:
         attributes = await get_genome_metadata(genome_id)
         annotation_provider_name = attributes.get("genebuild.provider_name")
-        annotation_version = attributes.get("genebuild.display_version")
-        last_updated_date = attributes.get("genebuild.last_geneset_update")
+        annotation_version = attributes.get("genebuild.display_version") or ""
```
azangru: didn't you change the attribute name?

azangru: Also, `.get` accepts a second argument which is the default return value

azangru (approved):

Mehrnaz-Charkhchi (approved):

PR #45, opened 2024-09-27
Update cicd pipeline (review deployment)

veidenberg:
### Description
Update GitLab CI/CD pipeline:
* fix the job for deploying review apps with custom Tools API backend
* add setup-review job for deploying review app dependencies
* update dev/main branch rules (dev->dev(default review)/internal/staging; main->prod)
* smaller fixes and edits (unify resource names, add ci/cd variables etc.)

### Related JIRA Issue(s)
[ENSWBSITES-2759](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2759)

### Review Apps
Full review app: http://fix-review-full.review.ensembl.org/vep
Pipelines: https://gitlab.ebi.ac.uk/ensembl-web/ensembl-client/-/pipelines/564282 , https://gitlab.ebi.ac.uk/ensembl-web/ensembl-tools-api/-/pipelines/564285
Backend-only: http://fix-review.review.ensembl.org/api/tools/blast/config
Pipeline: https://gitlab.ebi.ac.uk/ensembl-web/ensembl-tools-api/-/pipelines/564272

### Checklist
- [x] Yaml formatting
- [x] Tests

### Dependencies 
Related MRs: [k8s manifests MR](https://gitlab.ebi.ac.uk/ensembl-web/ensembl-k8s-manifests/-/merge_requests/43), [manifests commit](https://gitlab.ebi.ac.uk/ensembl-web/ensembl-k8s-manifests/-/commit/5cdebadcea6f35847475ec06fb397d859e4c8094), [k8s config MR](https://gitlab.ebi.ac.uk/ensembl-web/newsite-api-config-setup/-/merge_requests/21)
Added CI/CD variables to Tools API gitlab repo needed for the deploy jobs.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -135,7 +135,8 @@ Setup-review:
     - kubectl delete ingress ensembl-web-tools-api-ingress-${NS} || true
     - kubectl create namespace ${NS} || true
     - kubectl config set-context --current --namespace=${NS}
-    - kubectl create secret docker-registry ensemblweb-pull-secret --docker-server=${GITLAB_REGISTRY_URL} --docker-username=${WEB_PULL_DOCKER_USER} --docker-password=${WEB_PULL_DOCKER_SECRET} --docker-email=${WEB_PULL_DOCKER_EMAIL} || true
+    - kubectl create secret docker-registry ensemblweb-pull-secret --docker-server=${GITLAB_REGISTRY_URL} 
+      --docker-username=${WEB_PULL_DOCKER_USER} --docker-password=${WEB_PULL_DOCKER_SECRET} --docker-email=${WEB_PULL_DOCKER_EMAIL} || true
     # Setup ToolsAPI resources in review namespace (service, ingress, secret) + proxy config
     - git clone --depth 1 --branch k8s123-migration https://gitlab.ebi.ac.uk/ensembl-web/ensembl-k8s-manifests.git
     - cd ensembl-k8s-manifests/ensembl-web-tools-api
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -135,7 +135,7 @@ Setup-review:
     - kubectl delete ingress ensembl-tools-api-ingress-${NS} || true
     - kubectl create namespace ${NS} || true
     - kubectl config set-context --current --namespace=${NS}
-    - kubectl create secret docker-registry ensemblweb-pull-secret --docker-server=${GITLAB_REGISTRY_URL} \
+    - kubectl create secret docker-registry ensemblweb-pull-secret --docker-server=${GITLAB_REGISTRY_URL}
       --docker-username=${WEB_PULL_DOCKER_USER} --docker-password=${WEB_PULL_DOCKER_SECRET} --docker-email=${WEB_PULL_DOCKER_EMAIL} || true
     # Setup ToolsAPI resources in review namespace (service, ingress, secret)
     - git clone --depth 1 --branch k8s123-migration https://gitlab.ebi.ac.uk/ensembl-web/ensembl-k8s-manifests.git
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -94,7 +94,7 @@ Test:
   artifacts:
     when: always
     reports:
-      junit: report.xml
+      junit: app/report.xml
 
 # Build the docker image
 Docker:
@@ -135,8 +135,7 @@ Setup-review:
     - kubectl delete ingress ensembl-web-tools-api-ingress-${NS} || true
     - kubectl create namespace ${NS} || true
     - kubectl config set-context --current --namespace=${NS}
-    - "kubectl create secret docker-registry ensemblweb-pull-secret --docker-server=${GITLAB_REGISTRY_URL} \
-      --docker-username=${WEB_PULL_DOCKER_USER} --docker-password=${WEB_PULL_DOCKER_SECRET} --docker-email=${WEB_PULL_DOCKER_EMAIL} || true"
+    - kubectl create secret docker-registry ensemblweb-pull-secret --docker-server=${GITLAB_REGISTRY_URL} --docker-username=${WEB_PULL_DOCKER_USER} --docker-password=${WEB_PULL_DOCKER_SECRET} --docker-email=${WEB_PULL_DOCKER_EMAIL} || true
     # Setup ToolsAPI resources in review namespace (service, ingress, secret) + proxy config
     - git clone --depth 1 --branch k8s123-migration https://gitlab.ebi.ac.uk/ensembl-web/ensembl-k8s-manifests.git
     - cd ensembl-k8s-manifests/ensembl-web-tools-api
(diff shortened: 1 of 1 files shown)


kamaldodiya (changes requested):
1. I think having dedicated setup phase when the branch is created similar to Ensembl-client would be more appropriate solution.
2. Review job is failing - https://gitlab.ebi.ac.uk/ensembl-web/ensembl-tools-api/-/jobs/1821074

veidenberg:
@kamaldodiya 
> 1. I think having dedicated setup phase when the branch is created similar to Ensembl-client would be more appropriate solution.

In Ensembl-client the review setup step [deploys resources for other services](https://github.com/Ensembl/ensembl-client/blob/fix-review/gitlab-ci-templates/.setup-review-template.yaml) (default backends) which is not needed for Tools API (these resources have already been set up by Ensembl client). What would be the purpose of review setup here (and what would be then left to do for deploy-review step)? And should we also add setup review step for other backends that [have review step](https://github.com/Ensembl/ensembl-dauphin-style-compiler/blob/master/backend-server/.backend-server-ci.yml)?

> 2. Review job is failing - https://gitlab.ebi.ac.uk/ensembl-web/ensembl-tools-api/-/jobs/1821074

From the [linked job logs](https://gitlab.ebi.ac.uk/ensembl-web/ensembl-tools-api/-/jobs/1821074):
`namespaces "fix-review-test-1" not found. Please create 'fix-review-test-1' branch in ensembl-client and restart this pipeline`
This is due to the way we have set up our review apps deployment. Also see [this discussion in related PR](https://github.com/Ensembl/ensembl-web-tools-api/pull/42#discussion_r1773735438).
We could save a few clicks by automating the Ensembl client pipeline launch with e.g. [downstream pipelines](https://docs.gitlab.com/ee/ci/pipelines/downstream_pipelines.html), but the client pipeline will still be a dependency (because we need to setup the client and default backends for all review apps).

kamaldodiya:
> What would be the purpose of review setup here (and what would be then left to do for deploy-review step)?

Well Setup is a phase which would run the things which needs to be run just once to setup and not on every push. It's a requirment for Ensembl-client to setup non-default backend so it is in setup not necessarily every  application has setup requirement. Here it could be setting up configmap, secrets, service, ingress etc. 
for example https://github.com/Ensembl/ensembl-web-tools-api/blob/fix-review/.gitlab-ci.yml#L133-L137 this doesn't need to be applied for every push it is just needed once.

.gitlab-ci.yml
```diff
@@ -21,20 +21,35 @@ variables:
...
+.branch-rules-dev:
+  rules:
+    - if: '$CI_DEPLOY_FREEZE == null && $CI_COMMIT_BRANCH == "dev"'
+      when: on_success
+    - when: never
 
-.deploy-branch-rules:
+.branch-rules-review:
   rules:
     - if: '$CI_DEPLOY_FREEZE == null && ($CI_COMMIT_BRANCH != "main" || $CI_COMMIT_BRANCH != "dev")'
```
azangru: is `CI_DEPLOY_FREEZE` relevant for review deployments?

.gitlab-ci.yml
```diff
@@ -50,13 +65,22 @@ stages:
...
+.deploy-main:
+  extends:
+    - .deploy
+    - .branch-rules-main
+
+# Run tests
 Test:
   stage: test
+  extends:
+    - .deploy-rules
```
azangru: I am probably missing something; but why should the Test job be concerned with whether or not something needs to be deployed (and whether or not there is a deploy freeze)?

.gitlab-ci.yml
```diff
@@ -70,13 +94,13 @@ Test:
...
-      junit: report.xml
-  except:
-    - /^skiptest\/.*$/
+      junit: app/report.xml
 
 # Build the docker image
 Docker:
   stage: build
+  extends:
+    - .deploy-rules
```
azangru: Does building a Docker image need to be concerned with deployment rules?

.gitlab-ci.yml
```diff
@@ -88,86 +112,120 @@ Docker:
...
-    - kubectl -n ${NS}  set env deployment/ensembl-tools-api-deployment --from configmap/ensembl-tools-api-configmap
 
-# Stop feature branch review app (cleanup resources) 
+# Stop review app (cleanup backend resources)
 Stop_review:
-  extends: .deploy-branch-rules
   stage: deploy
   image: dockerhub.ebi.ac.uk/ensembl-web/deploy-tools:latest
+  rules:
+    - if: '$CI_DEPLOY_FREEZE == null && ($CI_COMMIT_BRANCH != "main" || $CI_COMMIT_BRANCH != "dev")'
```
azangru: Why does this job care about CI_DEPLOY_FREEZE?

.gitlab-ci.yml
veidenberg: To which env should deploy freeze mode apply to @kamaldodiya?

.gitlab-ci.yml
veidenberg: It depends whether we want to run the code tests when we know it's not going to be deployed downstream.

.gitlab-ci.yml
veidenberg: Should we build the docker image when it's not going to be used (deploy_freeze in place)?

.gitlab-ci.yml
veidenberg: True, it doesn't (as long as the review app was deployed).

.gitlab-ci.yml
azangru: I would have thought we want to run tests on every push regardless of where it goes, to get early feedback. At least this is how it is for the client.

.gitlab-ci.yml
kamaldodiya: It is not environment specific. if DEPLOY_FREEZE is set it would not allow that job to be scheduled - here deploy job to whatever (review) environment .

veidenberg: @kamaldodiya what I/@azangru meant is: should we follow `DEPLOY_FREEZE` in all environment deployment jobs (including review) or do we want to ignore it for certain deployment envs? For example ensembl-client [seems to only use it for production and dev](https://github.com/Ensembl/ensembl-client/blob/dev/.gitlab-ci.yml) deployments (but not for staging/internal/review), so we should probably follow the same policy here (and in all the other repos)? Otherwise we get broken (partly-deployed) site in some envs when `DEPLOY_FREEZE` is switched on.

azangru: My understanding was that `DEPLOY_FREEZE` only makes sense for us for production, or any other "important" deployment that for some reason should be prevented from updating immediately. Dev branch is on its way out; and feature review deployments (as well as probably staging?) should be updated immediately.

veidenberg: Removed freeze check from review/staging.

.gitlab-ci.yml
veidenberg: Removed freeze check from build-docker.

PR #48, opened 2024-10-01
Update error messaging in Tools API (VEP)

veidenberg:
### Description
Update error handling in VEP endpoints to help debugging the error root cause.

Handling flow:
1) Exceptions bubble from utility functions up to the caller (functions serving VEP endpoints). When it can, utility functions catch and add context to the error messages from upstream service before reraising.
2) Exceptions are handled by the caller depending of the exception type:
Exceptions with error messages from downstream services (Seqera cloud, metadata API) are logged and status code forwarded to the client. Internal errors are logged and 500 status sent to the client. Also added some context (caller name, error type) to the logged message.

### Related JIRA Issue(s)
[ENSWBSITES-2742](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2742)

### Review App URL(s) 
http://vep-errormsg.review.ensembl.org

### Examples

request: `api/tools/vep/submissions/2v7xiMUta5yOjM/status`
response: `200  {"workflow": {"id": "2v7xiMUta5yOjM", ...`
logs (docker output): (error content coming from `errorMessage` field in Seqera cloud response)
```bash
tools-api  | <timestamp> | ERROR | vep.vep_resources:vep_status:103 - VEP submission f2v7xiMUta5yOjM failed: f[W::bcf_hdr_parse] Could not parse header line: "#CHROM  POS     ID      REF     ALT     QUAL    FILTER  INFO"
tools-api  | [E::bcf_hdr_parse] Could not parse the header, sample line not found
tools-api  | Failed to read from interesting_single.vcf.gz: could not parse header
```
==============
request: `api/tools/vep/submissions/4mDEX0zwKr7RPL/status`
response: `200  {"workflow": {"id": "4IrKtDCJRy1XQJ", ...`
logs: (error content coming from `errorReport` field in Seqera cloud response)
```bash
tools-api  | <timestamp> | ERROR | vep.vep_resources:vep_status:103 - VEP submission f4mDEX0zwKr7RPL failed: 
<more lines>
tools-api  | ERROR ~ The specified input does not exist: <PATH>/small.vcf
tools-api  |  -- Check 'nf-4mDEX0zwKr7RPL.log' file for details
```
==============
request: `POST api/tools/vep/submissions` (with invalid genome ID)
response: `404 {"status_code": 404, ...`
logs: (error content coming from metadata API: get_vep_support_location)
```bash
tools-api  | <timestamp> | ERROR | vep.vep_resources:submit_vep:86 - VEP submission error (upstream): Could not find details for genome a7335667-93e7-11ec-a39d-005056b38 and dataset genebuild.
```
==============
request: `/api/tools/vep/form_config/<invalid_genome_id>`
response: `404  {"status_code": 404, ...`
logs: (error content coming from metadata API: get_genome_metadata)
```bash
tools-api  | <timestamp> | ERROR | vep.vep_resources:get_form_config:243 - VEP form config error (upstream): 404 Client Error: Not Found for url: https://beta.ensembl.org/api/metadata/genome/<genome_id>/dataset/genebuild/attributes?attribute_names=genebuild.provider_name&attribute_names=genebuild.display_version&attribute_names=genebuild.last_geneset_update
```

### Checklist

- [x] Black formatting
- [x] Tests

### Dependencies 
None

### DIFF
--- a/app/vep/models/pipeline_model.py
+++ b/app/vep/models/pipeline_model.py
@@ -101,5 +101,7 @@ class PipelineStatus(BaseModel):
     def serialize_status(self, status: str):
         if status == "UNKNOWN":
             status = "FAILED"
-            logging.info(f"Unknown status was returned for submission {self.submission_id}")
+            logging.info(
+                f"Unknown status was returned for submission {self.submission_id}"
+            )
         return status
--- a/app/vep/models/pipeline_model.py
+++ b/app/vep/models/pipeline_model.py
@@ -88,7 +88,6 @@ fasta {self.fasta}
                 ini_file.write(config_ini)
             return ini_file
         except Exception as e:
-            logging.info(e)
             raise RuntimeError("Could not create vep config ini file")
 
 
@@ -102,5 +101,5 @@ class PipelineStatus(BaseModel):
     def serialize_status(self, status: str):
         if status == "UNKNOWN":
             status = "FAILED"
-            logging.info("UNKNOWN STATUS WAS RETURNED HERE")
+            logging.info(f"Unknown status was returned for submission {self.submission_id}")
         return status
--- a/app/vep/models/pipeline_model.py
+++ b/app/vep/models/pipeline_model.py
@@ -1,3 +1,4 @@
+import logging, os
 from pydantic import (
     BaseModel,
     DirectoryPath,
@@ -7,13 +8,13 @@ from pydantic import (
     AliasPath,
     field_serializer,
 )
+from requests import HTTPError
+
 from core.config import (
     NF_COMPUTE_ENV_ID,
     NF_PIPELINE_URL,
 )
 from core.logging import InterceptHandler
-import logging, os
-
 from vep.utils.web_metadata import get_vep_support_location
 
 logging.getLogger().handlers = [InterceptHandler()]
@@ -87,8 +88,10 @@ fasta {self.fasta}
             with open(os.path.join(dir, "config.ini"), "w") as ini_file:
                 ini_file.write(config_ini)
             return ini_file
+        except HTTPError:
+            raise
         except Exception as e:
-            raise RuntimeError("Could not create vep config ini file")
+            raise RuntimeError(f"Could not create VEP config ini file: {e}")
 
 
 class PipelineStatus(BaseModel):
--- a/app/vep/models/pipeline_model.py
+++ b/app/vep/models/pipeline_model.py
@@ -88,8 +88,6 @@ fasta {self.fasta}
             with open(os.path.join(dir, "config.ini"), "w") as ini_file:
                 ini_file.write(config_ini)
             return ini_file
-        except HTTPError:
-            raise
         except Exception as e:
             raise RuntimeError(f"Could not create VEP config ini file: {e}")
 
--- a/app/vep/models/submission_form.py
+++ b/app/vep/models/submission_form.py
@@ -1,10 +1,8 @@
-from pydantic import BaseModel
-
+from pydantic import BaseModel, Field
 
 class DropdownOption(BaseModel):
-    label: str
-    value: str
-
+  label: str
+  value: str
 
 class Dropdown(BaseModel):
     label: str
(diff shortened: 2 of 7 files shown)


kamaldodiya (changes requested):
It would be good to see the actual error messages in logs. I tried to login to the review app but it seems it didn't deploy the pod. Also ingress is also missing so review app is still using the default tools-api-backend. Can you please fix this ? or else add the logs from your local run in the PR :)

app/vep/models/pipeline_model.py
```diff
@@ -7,13 +8,13 @@
...
     field_serializer,
 )
+from requests import HTTPError
+
 from core.config import (
     NF_COMPUTE_ENV_ID,
     NF_PIPELINE_URL,
 )
 from core.logging import InterceptHandler
-import logging, os
```
Mehrnaz-Charkhchi, line 15: Isn't this like causing issues on line 19?

app/vep/models/pipeline_model.py
```diff
@@ -87,9 +88,10 @@ def create_config_ini_file(self, dir):
             with open(os.path.join(dir, "config.ini"), "w") as ini_file:
                 ini_file.write(config_ini)
             return ini_file
+        except HTTPError:
```
Mehrnaz-Charkhchi: Why are we raising HTTP errors in file open action?

Mehrnaz-Charkhchi:
It would have been better if you could commit your changes first and then run black formatting!

app/vep/models/pipeline_model.py
veidenberg: Just reordering (see line 1)

app/vep/models/pipeline_model.py
veidenberg: Sorry, not needed here, removed it (was meant for forwarding httperrors from line 72).

veidenberg:
> It would have been better if you could commit your changes first and then run black formatting!

True. There's a separate black formating commit but some of it also went to the first commit, sorry about that.
However, the formatting noise is only an issue when it has not been used consistently (together with every change). Now that these files have been (re)formatted, future PRs should to be cleaner without the need to separate formatting commits.

Mehrnaz-Charkhchi:
> > It would have been better if you could commit your changes first and then run black formatting!
> 
> True. There's a separate black formating commit but some of it also went to the first commit, sorry about that. However, the formatting noise is only an issue when it has not been used consistently (together with every change). Now that these files have been (re)formatted, future PRs should to be cleaner without the need to separate formatting commits.

This isn't how commits work in github. there is now way for me to be able to review this PR properly without the noise!
I'm not against running black and we should do that often after our development. But we need to have manageable way to commit the code!

veidenberg:
> It would be good to see the actual error messages in logs. I tried to login to the review app but it seems it didn't deploy the pod. Also ingress is also missing so review app is still using the default tools-api-backend. Can you please fix this ? or else add the logs from your local run in the PR :)

Example outputs added to description.

kamaldodiya (approved):

app/vep/models/upload_vcf_files.py
```diff
@@ -92,8 +92,6 @@ async def stream(self):
         except (MaxBodySizeException, ValidationError):
             shutil.rmtree(self.temp_dir)
             raise MaxBodySizeException(MAX_FILE_SIZE)
-        except (Exception, ClientDisconnect) as e:
-            print(e)
-            logging.info(e)
+        except (ClientDisconnect, Exception):
```
Mehrnaz-Charkhchi, line 95: Could you please add the logging here?

veidenberg: This exception is already being logged by the caller ([on this line](https://github.com/Ensembl/ensembl-web-tools-api/blob/14cf9ea0c002388cfa2b5b92c4d4b13b0b903d72/app/vep/vep_resources.py#L91)), so double-logging it here would only add noise. Is there a benefit/purpose for adding logging here or would it be better to add this in the future when such a need arises? And in case we add logging here, should we also add it to the `MaxBodySizeException`/`ValidationError` above (also, I was wondering why is `ValidationError` re-raised as `MaxBodySizeException`)?

app/vep/models/submission_form.py
```diff
@@ -11,17 +13,21 @@ class Dropdown(BaseModel):
...
     symbol: Checkbox = Checkbox(label="Gene symbol")
     biotype: Checkbox = Checkbox(label="Transcript biotype")
 
+
 class GenomeAnnotationProvider(BaseModel):
-  annotation_provider_name: str
-  annotation_version: str
+    annotation_provider_name: str
+    annotation_version: str
+    last_geneset_update: str
```
Mehrnaz-Charkhchi: what is this update for?

veidenberg: I added a missing property to the datamodel (`last_geneset_update`) but there are more issues with this model and the related code so I reverted my chagnes in this file (not needed for this PR anyway) and will open a separate PR to to address it. Also, by taking the file out of this PR I reduced the number of Black-modified lines to just 10 (out of ~140 added lines) so I hope the formatting noise is manageable now.

PR #50, opened 2024-10-04
Update Tools API openAPI contract and regenerate VCF results pydantic model

Mehrnaz-Charkhchi:
### Description
Update Tools API openAPI contract and regenerate VCF results pydantic model.

### Related JIRA Issue(s)
[ENSWBSITES-2781](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2781)

### Review App URL(s) 

### Checklist

- [ ] Black formatting
- [ ] Tests


app/tests/models/test_vcf_results_model.py
```diff
@@ -0,0 +1,164 @@
...
+)
+
+
+class TestVCFResultModel(unittest.TestCase):
+
+    def test_predicted_intergenic_consequence(self):
+        consequence = PredictedIntergenicConsequence(
+            feature_type=None,
+            consequences=["intergenic_variant"]
+        )
```
azangru: Seeing now for the first time how the `PredictedIntergenicConsequence` class is instantiated, I think passing both these arguments to it is redundant (it must be the result of it being auto-generated from the schema). If the contract is that the `PredictedIntergenicConsequence` class will always serialise to:

```json
{
  "feature_type": null,
  "consequences": ["intergenic_variant"]
}
```

and if the caller of the class always knows that it is calling the `PredictedIntergenicConsequence` class, then there shouldn't be any need to pass any data into it. This class can have the `feature_type` field set always to `None`,  and the `consequences` field set always to `["intergenic_variant"]`.

jyothishnt (approved):

azangru (approved):

PR #53, opened 2024-10-25
Support path update

Mehrnaz-Charkhchi:
### Description
Production want to split the VEP support dataset attribute vep.bgz_location into two entries, one for GFF vep.gff_location and one for FASTA vep.faa_location. Update Tools API's get_vep_support_location method to support these two new attributes

### Related JIRA Issue(s)
[ENSWBSITES-2802](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2802)

### Review App URL(s) 

### Checklist

- [ ] Black formatting
- [ ] Tests

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -32,6 +32,8 @@ stages:
   rules:
     - if: $CI_DEPLOY_FREEZE == null && $CI_COMMIT_BRANCH == "main"
       when: on_success
+    - when: never
+
 
 .branch-rules-review:
   rules:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -31,8 +31,6 @@ stages:
 .branch-rules-main:
   rules:
     - if: $CI_DEPLOY_FREEZE == null && $CI_COMMIT_BRANCH == "main"
-      when: on_success
-    - when: never
 
 
 .branch-rules-review:
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -31,7 +31,7 @@ stages:
 .branch-rules-main:
   rules:
     - if: $CI_DEPLOY_FREEZE == null && $CI_COMMIT_BRANCH == "main"
-
+      when: on_success
 
 .branch-rules-review:
   rules:
--- a/app/vep/models/pipeline_model.py
+++ b/app/vep/models/pipeline_model.py
@@ -71,8 +71,8 @@ class ConfigIniParams(BaseModel):
 
     def create_config_ini_file(self, dir):
         vep_support_location = get_vep_support_location(self.genome_id)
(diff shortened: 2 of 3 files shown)


azangru (approved):
Code looks good; but I think the actual paths to files returned from grpc may be incorrect. Not sure what this means for the PR. I think the moment production-1 database is copied over to our database it will break VEP anyway; so probably, the wrong paths from grpc shouldn't be a blocker for this PR.

.gitlab-ci.yml
```diff
@@ -31,8 +31,6 @@ stages:
 .branch-rules-main:
   rules:
     - if: $CI_DEPLOY_FREEZE == null && $CI_COMMIT_BRANCH == "main"
-      when: on_success
```
kamaldodiya: Removing this would cause the deploy job to schedule/run even if the jobs in earlier stages will fail. This would also make the deployment to fail even though job would succeed. This was the safeguard to prevent deployment failure. 
Refer gitlab docs here https://docs.gitlab.com/ee/ci/yaml/#when

kamaldodiya (changes requested):

.gitlab-ci.yml
Mehrnaz-Charkhchi: @kamaldodiya My issue was that live deployments deployed automatically instead of waiting for manual deployment I checked another project and found they don't have these two lines.

kamaldodiya: > My issue was that live deployments deployed automatically instead of waiting for manual deployment

That is perfectly fine but the thing is that it shouldn't have been scheduled/run if th build fails. That is what on_success does 

>  I checked another project and found they don't have these two lines.

Ensembl-client has different mechanism to safe guard against such scenario  ( we use needs in Ensembl-client ) as it has different requirement where both static assets and node build has to finish  before it can trigger deployment and   has to be deployed together which is why we use needs ( as it provides such functionality)

Mehrnaz-Charkhchi: is adding  `when: on_success` enough?

veidenberg: @Mehrnaz-Charkhchi I couldn't find it in the docs, but [it seems that](https://stackoverflow.com/questions/75852822/gitlab-ci-is-there-a-difference-between-a-rule-with-whenalways-to-a-rule-with) `rules:when:` overrides `when:` value. Therefore removing `rules:when: on_success` from `.branch-rules-main` template should work (so `when: manual` will be used in production jobs and `when: on_success` (implicit) in staging/dev jobs).

veidenberg: @kamaldodiya adding `rules:when: on_success` should not be needed ([default value](https://docs.gitlab.com/ee/ci/yaml/#ruleswhen)). Also we cannot have `on_success` for `manual` jobs (mutually exclusive values for `when`).

PR #54, opened 2024-11-22
Integrate bcftools

veidenberg:
### Description
Integrate command-line based [bcftools ](http://samtools.github.io/bcftools/) to the VEP results page endpoint.

### Related JIRA Issue(s)
[ENSWBSITES-2727](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2727)

### Review App URL(s) 
http://integrate-bcftools.review.ensembl.org

### Knowledge Base
[Bcftools docs](http://samtools.github.io/bcftools/howtos/query.html) for working with VCF files.
[Related JIRA](https://www.ebi.ac.uk/panda/jira/browse/ENSWBSITES-2714) for more background.

### Example(s)
Run VEP for a large input file, check for delays when browsing the results table.
Run unit tests: `cd app; python -m pytest tests/test_vep.py`

### Checklist

- [x] Black formatting
- [x] Tests

### Dependencies 
None


app/vep/utils/vcf_results.py
```diff
@@ -66,13 +66,15 @@ def _get_prediction_index_map(
         if csq_headers[index] in target_columns
     }
 
-    # TODO add exception if len(map) != len(TARGET_COLUMNS)
+    #if len(map_index) != len(target_columns): #CANONICAL is often missing
+    #    raise Exception(f"Required columns missing from CSQ header: {map_index.keys()}")
```
azangru: Are these two commented lines left intentionally, or forgotten?

app/vep/utils/vcf_results.py
```diff
@@ -119,9 +121,9 @@ def _get_alt_allele_details(
             consequences.append(
                 model.PredictedTranscriptConsequence(
                     feature_type=model.FeatureType.transcript,
-                    stable_id=_get_csq_value(csq_values, "Feature", None, index_map),
-                    gene_stable_id=_get_csq_value(csq_values, "Gene", None, index_map),
-                    biotype=_get_csq_value(csq_values, "BIOTYPE", None, index_map),
+                    stable_id=_get_csq_value(csq_values, "Feature", "", index_map),
+                    gene_stable_id=_get_csq_value(csq_values, "Gene", "", index_map),
+                    biotype=_get_csq_value(csq_values, "BIOTYPE", "", index_map),
                     is_canonical=is_cononical,
```
azangru: `is_cononical ` — check where this variable is used, and change `cononical` to `canonical`

app/vep/utils/vcf_results.py
```diff
@@ -144,29 +146,70 @@ def _get_alt_allele_details(
     )
 
 
+def _get_meta(vcf_path: FilePath) -> model.VcfMetadata:
```
azangru: If, as the comment says, this is a method for managing vcf file metadata, then perhaps its name should be `_get_vcf_file_meta`, or something of the sort?

app/vep/utils/vcf_results.py
veidenberg: Itentional (kind of). I implemented the TODO comment, but it turned out to be a bad idea: a missing csq column in the results file (like "canonical") raised the exception although it's recoverable (by the default values defined in downstream code). The fate of this these lines needs a decision, but it's not in scope of this PR so keeping the comments for reference.

app/vep/utils/vcf_results.py
veidenberg: It's also meant to contain other information (e.g. filter selection status) but that's not implemented yet so renaming it for now.

kamaldodiya (approved):

jonk-ebi:
The unit tests are failing, this is due to the update to `_get_alt_allele_details` where the reference is now passed. Also `get_results_from_stream` now needs total, will comment about that again in a moment

You can test them by running `python -m pytest tests/test_vep.py` from the `app` directory. 

Test output 
```
FAILED tests/test_vep.py::test_get_alt_allele_details - TypeError: _get_alt_allele_details() missing 1 required positional argument: 'index_map'
FAILED tests/test_vep.py::test_get_alt_allele_no_consequence - TypeError: _get_alt_allele_details() missing 1 required positional argument: 'index_map'
FAILED tests/test_vep.py::test_get_alt_allele_details_intergenic - TypeError: _get_alt_allele_details() missing 1 required positional argument: 'index_map'
FAILED tests/test_vep.py::test_get_results_from_stream - TypeError: get_results_from_stream() missing 1 required positional argument: 'vcf_stream'
FAILED tests/test_vep.py::test_paging - TypeError: get_results_from_stream() missing 1 required positional argument: 'vcf_stream'
FAILED tests/test_vep.py::test_negative_paging - TypeError: get_results_from_stream() missing 1 required positional argument: 'vcf_stream'
```

app/vep/utils/vcf_results.py
```diff
@@ -144,117 +147,131 @@ def _get_alt_allele_details(
...
-    APISpecification.yaml for a given VCF stream"""
+    """Helper method to convert a filestream to VCF records for _get_results_from_vcfpy"""
 
     # Load vcf
     vcf_records = vcfpy.Reader.from_stream(vcf_stream)
-    return _get_results_from_vcfpy(page_size, page, vcf_records)
-
+    return _get_results_from_vcfpy(page_size, page, total, vcf_records)
 
+import sys
```
jonk-ebi: This should be with the other imports

app/vep/utils/vcf_results.py
```diff
@@ -144,117 +147,131 @@ def _get_alt_allele_details(
...
+    page_size = min(max(page_size, 0), total)
+    row_offset = min(page * page_size, total) + vcf_info.header_count
+    vcf_headers = subprocess.check_output( # fetch all header rows
+        f"bcftools view -h {vcf_path}", shell=True, text=True
+    )
+    vcf_slice = subprocess.check_output( # fetch subset of variant rows
+        f"bcftools view {vcf_path} | head -n{row_offset} | tail -n{page_size}",
+        shell=True, text=True
+    )
+    vcf_stream = StringIO(vcf_headers + vcf_slice)
```
jonk-ebi, line 204: Ideally this should be just in `_get_results_from_vcfpy`, get_results_from_path and get_results_from_stream are just public interfaces to the `_get_results_from_vcfpy` 

EDIT
Ah I see, `_get_results_from_vcfpy` works from a stream but the meta function uses sub processes that operate on files. 

The streaming is used to support embedded vcf test data in the unit tests. It might be worth rethinking the tests given that `app` doesn't resolve from pytest

app/vep/utils/vcf_results.py
```diff
@@ -144,117 +147,131 @@ def _get_alt_allele_details(
     )
 
 
+def _get_vcf_meta(vcf_path: FilePath) -> model.VcfMetadata:
```
jonk-ebi, line 151: A lot can go wrong here. It would be good to add some validation or throw some meaningful errors.
- bcftools missing
- bcftools output differs slightly from expected - I see it is locked in the docker so probably fine
- variant_count assumes multiple `:` and that the last element on the split is a number
- grep's output can different on different shells, distros - I think you are safe with what you are doing

jonk-ebi (commented):
Looks good. Normally when I use subprocess I capture the errout for reporting on the error state. Not sure how that works with `check_output` as the docs say it raises an exception when it gets a non zero exit code

app/vep/utils/vcf_results.py
veidenberg: Most runtime failures in this python file are not recoverable (data issues), which will throw an exception, resulting in a 500 response and a line in error log. I've added some more context to exception messages, hopefully it helps with debugging. For the mentions:
* bcftools is installed in dockerfile (passed image build test). Failures from running bcftools will throw `CalledProcessError` (with message like `<command> returned non-zero exit status` logged).
* bcftools output is checked by datamodel, e.g. non-ints in the count variables will throw KeyError
* the full line from `bcftools stats | grep` looks like `SN 0 number of records: 12345` (unique keywords + a colon expected). Anything else should throw an exception.

app/vep/utils/vcf_results.py
veidenberg: Debug code, removed

app/vep/utils/vcf_results.py
veidenberg: Yes, I noticed the stream consumer was only used in tests. However, the single-page vcf slice from bcftools is also a stream (no need to store it to a file), so the flow is now `get_results_from_path` > `get_results_from_stream` > `_get_results_from_vcfpy`.  Tests need some work (including input files).

jonk-ebi (approved):
Code changes look good. It would be great to get the tests working and extend to include missing CSQ keys and a test for zero variants

jonk-ebi:
Tests are now mostly working for vep.

PR #56, opened 2025-07-15
Fix VEP file path

bilalebi:
### Description
Fix VEP file path

Slack discssion: https://genomes-ebi.slack.com/archives/C032E9YJ2N9/p1752489212371539


veidenberg (approved):
Looks good to me.
For context, `with_suffix` will truncate the base filename if it contains a dot, which is not what we want. 
New code tested in python REPL.

PR #58, opened 2026-06-12
Update DB path for BLAST by adding the genome uuid prefix

bilalebi:
### Description

Since it's a tiny change, I went a head and updated the DB path for BLAST by adding the genome UUID prefix.

Please have a look and make sure I didn't miss (up) anything

### DIFF
--- a/app/blast/blast.py
+++ b/app/blast/blast.py
@@ -106,7 +106,8 @@ suffix_map = {"dna": "unmasked", "dna_sm": "softmasked"}  # dna/dna_sm/cdna/pep
 
 
 def get_db_path(genome_id: str, db_type: str) -> str:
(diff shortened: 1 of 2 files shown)


veidenberg (approved):

app/blast/blast.py
```diff
@@ -106,7 +106,8 @@ class JobIDs(BaseModel):
 
 
 def get_db_path(genome_id: str, db_type: str) -> str:
-    return f"ensembl/{genome_id}/{suffix_map.get(db_type, db_type)}"
+    genome_id_prefix = genome_id[:3]
+    return f"ensembl/{genome_id_prefix}/{genome_id}/{suffix_map.get(db_type, db_type)}"
```
azangru, line 110: Is this something that the tools api has to work out on its own, or something that it should ask metadata about?

azangru (approved):

